PR #52, opened 2013-12-17
Close #51 - Update front-end dependencies

### LINKED ISSUE #51: Update dependencies
Could this project's front-end dependencies be updated ? Particularly the FontAwesome set which – in the version in use – uses generic class names that clash with applications the debug bar is installed on. 

Ideally I think this project's dependencies should be managed by [Bower](http://bower.io) in order for them to be kept up to date.

The same thing can be said about the jQuery dependency that is now more than two years (!!) old.

Anahkiasen:
This moves managing of the front-end assets to Bower and update the existing dependencies to avoid icon conflicts.

### DIFF
--- a/src/DebugBar/DataCollector/TimeDataCollector.php
+++ b/src/DebugBar/DataCollector/TimeDataCollector.php
@@ -182,7 +182,7 @@ class TimeDataCollector extends DataCollector implements Renderable
     {
         return array(
             "time" => array(
-                "icon" => "time",
+                "icon" => "clock-o",
                 "tooltip" => "Request Duration",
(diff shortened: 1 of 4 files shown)


glensc:
how does this exactly work?

what invokes population of `src/DebugBar/Resources/vendor` during composer install?

barryvdh:
That folder is populated with `bower install`. The result is committed to the git repo, so no need to do anything during composer install.

glensc:
so .bowerrc and bower.json can be added to export-ignore in .gitattributes?

barryvdh:
Yes

Issue #87, opened 2014-02-04
Clear old storage automatically


barryvdh:
Would it be a good idea to clear the storage after a while? The collectors would keep on heaping up if you forget to do it by hand.
I created a simple garbage collection method that runs every few requests that removes all files older then 24 hours. Would this be handy for all storages?
See https://github.com/barryvdh/laravel-debugbar/commit/5aa061810042f432234ee062c7e2c01c9626cdc1

PR #89, opened 2014-02-12
(WIP) Add GarbageCollector

barryvdh:
Initial commit, created BaseStorage with gc functions, updated the interface.
Not really sure what the best place to run the GC would be..
Could be in the OpenHandler->handle() but that is only run on ajax/session requests. Could also call the checkGc() function after every save. (See #87)

Note: This is work in progress, not ready to merge yet..

### DIFF
--- a/src/DebugBar/OpenHandler.php
+++ b/src/DebugBar/OpenHandler.php
@@ -19,6 +19,7 @@ class OpenHandler
 
     /**
      * @param DebugBar $debugBar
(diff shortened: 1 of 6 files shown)


ilumos:
What's the latest on this?

I've just switched to the PDO storage driver, as using headers is getting too large for Nginx, and the file driver is creating so many JSON files that Vagrant is beginning to struggle.

In any case, loving the debugbar.

PR #187, opened 2014-12-23
Timedata memory usage

barryvdh:
Next to measuring the time, it might be interesting to keep track of the memory usage. Thoughts?

Tasks:
- [ ] Test better
- [ ] Tweak layout?


maximebf:
I didn't add that before because this is not accurate at all unfortunately. This reports only the delta between memory at start and end but not the actual usage during the interval. For example, setting and then unsetting a variable during this interval would not be reported by this.

spotman:
@maximebf @barryvdh This is a feature I need very much. Maybe `memory_get_peak_usage(true)` can be used instead or in pair with `memory_get_usage(true)`? This can visualize memory usage spikes during operation.

PR #190, opened 2015-01-05
Add Util class

### LINKED ISSUE #189: Cannot use mb_check_encoding()
What can I do to make this function work on my centos 6.2 machine?

I have tried to install it using ```yum``` but it didn't work.

barryvdh:
This adds a Util class to provide common methods. This also checks if functions exists, because not every php installation has mbstring installed.
I also added static access to the DataFormatter class, to provide some easy access in case anyone wants to use that.

Should fix #189 by checking mbstring functions.

//cc @maximebf for approval

### DIFF
--- a/src/DebugBar/Util.php
+++ b/src/DebugBar/Util.php
@@ -1,0 +1,130 @@
+<?php
+/*
+ * This file is part of the DebugBar package.
+ *
+ * (c) 2014 Maxime Bouroumeau-Fuseau
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace DebugBar;
+
+use DebugBar\DataFormatter\DataFormatter;
+use DebugBar\DataFormatter\DataFormatterInterface;
+
+/**
+ * Debugbar Utility Helper
+ *
+ * Helps provide common functions with a static interface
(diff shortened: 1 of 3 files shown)


src/DebugBar/Util.php
```diff
@@ -0,0 +1,130 @@
...
+    public static function formatDuration($seconds)
+    {
+        return self::getDataFormatter()->formatDuration($seconds);
+    }
+
+    /**
+     * Transforms a size in bytes to a human readable string
+     *
+     * @param  string $size
+     * @param  integer $precision
```
GrahamCampbell: should be int

barryvdh: Ah yes I just copied them from the interface (which has a few tiny mistakes also)

maximebf:
I'm ok with it

src/DebugBar/Util.php
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * <code>
+ *     if (!Util::isUtf8($string)) {
+ *          $string = Util::toUtf8($string);
+ *     }
+ * </code>
+ */
+class Util
+{
+
```
GrahamCampbell: shouldn't be this extra line here.

PR #196, opened 2015-02-05
Show the real parameter name for prepared queries.

hieblmedia:
Array index works only with positioned (where foo = ?) parameters.
To make named (where foo = :baz) parameters debuggable the name must be defined as key.


barryvdh:
Thanks

PR #202, opened 2015-03-11
update readme

snowair:
add phalcon debugbar


barryvdh:
Thanks.

Perhaps a readme in english would be nice though.

PR #203, opened 2015-03-17
baseUrl calculated from DOCUMENT_ROOT

johnc1984:
Composer files are not always in the root of the docroot in some projects. Should the DOCUMENT_ROOT not be available it reverts to the hard coded path. This method I find quicker than having to declare the path when calling the DebugBar in new projects, or when the composer is updated.


barryvdh:
Can't you just pass the base URL as parameter?

johnc1984:
yes you can! Ha, looked over that twice! cheers

PR #206, opened 2015-03-25
Add a div at the end of body for a bottom offset

bobd91:
Adding CSS padding-bottom to the body element does not work in some cases  (body height set to 100% fills the viewport and content overflows right through the padding)

This solution adds a div after the debugbar and recomputeBottomOffset is modified to set the height of this div to the height of the debugbar.


barryvdh:
Why did you close this?

bobd91:
Ah, yes.  A bit of a mistake.  I was tidying up and deleted my branch which of course closed the pull request.

I can restore the branch but don't think that it is suitable to merge at present as it replaces the padding mechanism.  I don't really like the extra div as it is in the flow of the main page but can't find a better way for Magento as the padding does not work.  

Perhaps it should be an additional option rather than a replacement?

I am happy to restore the branch and modify it to create an additional option if you think it worthwhile.

PR #210, opened 2015-04-20
Show whitespace in log output

nsams:
This allows logging multiple whitespace and whitespace at line start.

With that you can create indendation eg. in Timeline.


barryvdh:
Thanks!

PR #211, opened 2015-04-20
Improve accuracy of Timeline

nsams:
Don't round left/width, instead use two decimals.

This improves accuracy of Timeline when more similar sized entries are shown.


barryvdh:
Thanks!

PR #216, opened 2015-06-03
Hiding PHPDebugBar on @media print

fernandosavio:
PHPDebugBar showing in `@media print` makes no sense as it's purpose is debug.
This makes easier to test printing pages on debug server.


barryvdh:
Thanks!

PR #217, opened 2015-06-07
Remove meaningless hrefs

kokokurak:
Links have `href` attribute set to `javascript:`, which is not only useless, but unnecessarily shows destination on hover. It's annoying for debugbar, because it takes some place on the screen when we just hover on any tab. HTML5 specification allows anchors with no `href` specified, as mentioned here: http://w3c.github.io/html-reference/a.html#a.attrs.href

The only difference is that they then get standard cursor by default, which is corrected as well by my pull merge.

### DIFF
--- a/src/DebugBar/Resources/debugbar.css
+++ b/src/DebugBar/Resources/debugbar.css
@@ -20,7 +20,8 @@ div.phpdebugbar {
   line-height: 1;
 }
 
-div.phpdebugbar a {
+div.phpdebugbar a,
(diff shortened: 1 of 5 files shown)


barryvdh:
Makes sense. Is this PR finished?

kokokurak:
Yes it is :)

barryvdh:
Thanks!

kokokurak:
You're welcome! Awesome tool.

PR #218, opened 2015-06-20
Enable caseless searching in the "Files" toolbar

### LINKED ISSUE #194: Make search case-insensitive
If someone could point me in the direction as to where in the js to look for the search bar input I would like to create a PR to do case insensitive searching (for example in the files collector tab).

It would be of great use not to have to remember what casing the filename is in.

tonglil:
Most of the time I want to see if a file was loaded and I don't really want to bother getting all of the camelCasing when I'm just doing a simple search.

This enables caseless matching in the files collector tab!

However, if caseless searching returns too many results, you can still search with Capitals and get a more narrowed set of results :smile: 

Closes #194.

### DIFF
--- a/src/DebugBar/Resources/widgets.js
+++ b/src/DebugBar/Resources/widgets.js
@@ -318,12 +318,12 @@ if (typeof(PhpDebugBar) == 'undefined') {
 
             this.bindAttr(['exclude', 'search'], function() {
                 var data = this.get('data'),
-                    exclude = this.get('exclude'), 
-                    search = this.get('search'),
+                    exclude = this.get('exclude'),
+                    search = this.get('search').toLowerCase(),
                     fdata = [];
 
                 for (var i = 0; i < data.length; i++) {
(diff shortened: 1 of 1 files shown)


tonglil:
What can I do to get this feature into the tool?

barryvdh:
So it's only caseless when you type your word in lowercase? Not sure if that makes total sense. I guess it would make more sense to just always search caseless

tonglil:
So with this change, it defaults to caseless unless a case specified.

Filename: TestClass.php

Searches:
Before:
```
- TestClass => found
// Too lazy to write (all of the) caps or don't know what is cap'd because file is badly named
- Testclass => not found
- testclass => not found
```

After:
```
- TestClass => found
// Too lazy to write (all of the) caps or don't know what is cap'd because file is badly named
- Testclass => found
- testclass => found
```

There are some other situations where you want to see if a specific file was called but you don't remember the filepath or etc (I don't have a specific example at the moment).

barryvdh:
Ok thanks!

PR #227, opened 2015-09-13
Fix for wrong namespace lookup. No clean use possible otherwise.

clickalicious:
This fixes a bug - when extending the DataFormatter its not possible to set the new class as data formatter via collectorInstance->setDataFormatter(...) - it complains about wrong Interface.


barryvdh:
Can you use 'use DebugBar\DataFormatter\DataFormatterInterface;` instead?

GrahamCampbell:
Yeh, it would definitely be the best idea to just import it.

clickalicious:
Yeh, changed to your recommendation. bit late yesterday :D Seems cleaner.

barryvdh:
Thanks!

PR #233, opened 2015-10-29
Update .gitattributes

glensc:
exclude bower files from composer install

see discussion here: https://github.com/maximebf/php-debugbar/pull/52#issuecomment-152299029


barryvdh:
Thanks

PR #245, opened 2016-01-05
Added PHP_SAPI in Php data collector

pavankumarkatakam:
It is important to know how the php is running from cli or a web request. In some frameworks like laravel there is commands features which we also want to know the exception from command line or a web interface.


barryvdh:
Thanks

PR #246, opened 2016-01-05
Added support for negative numbers in DataFormatter::formatBytes()

darwish:
It's possible for a `TracedStatement` to have a negative memory delta, which causes the DataFormatter to crash.


barryvdh:
Thanks!

PR #247, opened 2016-01-07
remove the jquery.min.map reference

kife-design:
Remove the reference since its not included. This results in a NotFoundHttpException


barryvdh:
Thanks!

PR #256, opened 2016-02-02
Use SQLLogger interface instead of concrete DebugStack to allow different implementations

eng1neer:


barryvdh:
So queries are retrieved using: `$this->debugStack->queries`. This isn't in the SQLLogger interface, so we don't know if other instances also provides this functionality, right? For example, the EchoSQLLogger doesn't have this..

eng1neer:
You're right, SQLLogger interface does not provide a way to get the queries back.

What I'm trying to achieve is to be able to pass an object of a different type than DebugStack. I already have a data collector and need to implement some SQLLogger-like interface without inheriting from DebugStack.

Maybe there's some other way to do this.

eng1neer:
Yeah, I guess, that the best thing is not to combine two data collectors into one, as I was trying to, but use them separately and subclass Doctrine's DebugStack.

Thanks, closing this.

PR #257, opened 2016-02-12
fixed XSS vulnerability in sql parameter value rendering

jeff1985:
We found a security issue in the javascript part of sql sqlqueries widget: 
The values of parameters are rendered in a html table without any escaping, which makes it possible to do an XSS attack.

One might say, the debugbar is not intended to be used in production. Even if you ignore the security aspect of this issue, we might have rendering errors, if the params have any html in them. 

What i did to test:
`$entityManager->createQuery("select a from \My\Model\Comments a where a.kommentar=:c")->setParameter("c", "<script>alert();</script>")->execute();` 

We are using doctrine in our project. This resulted in mysql query with one parameter. The value of the parameter in debugbar appeared to be empty. Instead i saw the javascript alert.

Easy to fix with an escaping function.

### DIFF
--- a/src/DebugBar/Resources/widgets/sqlqueries/widget.js
+++ b/src/DebugBar/Resources/widgets/sqlqueries/widget.js
@@ -1,7 +1,25 @@
 (function($) {
 
     var csscls = PhpDebugBar.utils.makecsscls('phpdebugbar-widgets-');
+    
+    var entityMap = {
+        '&': '&amp;',
+        '<': '&lt;',
+        '>': '&gt;',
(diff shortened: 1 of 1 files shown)


barryvdh:
Should we escape on the client or server side? Might be safer on the server side?

jeff1985:
You should escape, where you generate html from user contributed input. In this particular case html is beeing generated on the client side, so it would be appropriate to escape it on the client side. 

I also checked, if the same is happening with the messages widget: nope, here its all fine.

Issue #266, opened 2016-04-14
wrong direction when body is rtl


ooghry:
hi all 
if body direction is set to rtl, php-debugbar direction set to rtl too. and that is some clutter as you can see in attachments.
![php-debugbar1](https://cloud.githubusercontent.com/assets/200244/14535585/b9b0c47c-0283-11e6-8926-c77006439da8.PNG)
![php-debugbar](https://cloud.githubusercontent.com/assets/200244/14535586/b9c30678-0283-11e6-9838-f5b8a7db12d2.PNG)

barryvdh:
Can you submit a PR?

PR #267, opened 2016-04-15
fix direction in rtl page

ooghry:
#266 i try to fix this issue.


barryvdh:
Do you need this so often? Can't we just add it to `div.phpdebugbar`

PR #268, opened 2016-04-16
[WIP] Add resettable interface for Collectors

barryvdh:
This is aimed specifically for https://github.com/php-pm/php-pm but could be used for other concurrent request frameworks (which I don't know of).

Adds a `reset()` method to the collectors to reset the state. Eg. clear the messages, reset starting time etc.
Not all collectors have been done yet, not all need it.

This can be used like `$debugbar->reset()` to reset all collectors, the collected data and regenerate an request id.
For Laravel this can be called in the middleware, after handling the request. For PHP-PM, this avoids duplicate entries etc.


barryvdh:
This might be still relevant for Octane etc. also

PR #270, opened 2016-04-28
Use included Font Awesome for debug bar control buttons

tburry:
Replace the CSS bitmaps with Font Awesome for the minimize, maximize, close, and open buttons. When you are on a retina monitor the lower DPI buttons really stick out and since you already use Font Awesome it makes sense to use it here too.

### DIFF
--- a/src/DebugBar/Resources/debugbar.css
+++ b/src/DebugBar/Resources/debugbar.css
@@ -174,9 +174,10 @@ a.phpdebugbar-tab.phpdebugbar-active {
     color: white;
   }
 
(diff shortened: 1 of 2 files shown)


barryvdh:
The idea was that even without Font Awesome, the important buttons were still usable. You can (for example) disable some vendor files, which conflict.
Perhaps we could use inline SVG, to provide retina icons (like https://github.com/blog/2112-delivering-octicons-with-svg )

tburry:
If it's okay I'll go ahead and close this since #272 was merged. Re-open if you wish.

PR #272, opened 2016-05-07
Use SVG for icons instead of PNG

tburry:
As an alternative to #270 based on the suggestion in [this comment](https://github.com/maximebf/php-debugbar/pull/270#issuecomment-215638893) I’ve provided a CSS file with embedded SVGs rather than the included font awesome.

There are pros and cons to each approach, but I’d love to see vector icons in the default debug bar.


barryvdh:
Thanks! So what are the cons of this approach?

tburry:
Here's what I'd list:

1. Browser support for SVG. This one is probably no big deal.
2. You can't style the background color of the icons as they are. This could be changed by adding the SVGs in the javascript file, but I figured that solution required a bit more that just hard-coded SVGs. As it stands, the current SVGs in the PR are no different than the previous PNGs in terms of functionality.
3. Using font awesome gives you more icons and makes it easier to use them next to text without having to write a lot of extra CSS.
4. From an implementation standpoint, it felt very easy to add the font awesome stuff, even though I had to alter the javascript. It seemed like if your UI had a lot of icons it would be no big deal to add more. On the other hand, adding the SVGs was a bit of a process: export, optimize, manually cut out extra stuff, url encode, paste into CSS. It wasn't that difficult, but I wouldn't want to manually go through too many SVGs.

I've been working with SVGs a lot recently to try and come up with a system that's as easy enough to use in place of an icon font. There are a lot of great advantages to SVG, but they aren't a clear cut winner against icon fonts. Sometimes SVGs feel a little bit bleeding edge when you start to try and style them or align them with text. There's also a few different SVG embedding techniques, but not one "best" one. Icon fonts work one way.

That might be where my "pros and cons" statement came from. I think SVGs are really going to be a great thing in the future. I was just hoping they would be a 100% rainbows and unicorns scenario and they aren't.

I will say though that being able to add a new icon without having to recompile an entire font or knowing that your app isn't including the overhead of two versions of the Pied Piper logo is a huge point in favor of SVGs. I'm also hoping to see design trends take advantage of multi-color icons again once this SVG thing catches on.

barryvdh:
OK cool

barryvdh:
Using the svg path in the collectors would allow is to decouple from font awesome, which causes interoperability problems ( version mismatch etc)

Issue #264, opened 2016-04-05
ParseError PHP7


j-combee:
In php7 they introduced a new error class, ParseError http://php.net/manual/en/class.parseerror.php.
This error type conflicts with the ExceptionsCollector::addException(Exception $e);

I have fixed this myself with adding a instanceof check like this:

```
    /**
     * Adds an exception to be profiled in the debug bar
     *
     * @param Exception $e
     */
    public function addException(Exception $e)
    {
        $this->exceptions[] = $e;
        $previousError = $e->getPrevious();
        if ($previousError instanceof Exception) {
            if ($this->chainExceptions && $previous = $e->getPrevious()) {
                $this->addException($previous);
            }
        }
    }
```

I put this in a issue and not in a push request since i do not know if there are more new classes that should be considered or if this fix is a correct fix for this issue.

rossbearman:
I've just run into this issue myself. As all exceptions and errors in PHP 7 inherit from the Throwable class, I changed the signatures of `addException(Exception $e)` and `formatExceptionData(Exception $e)` to `addException(Throwable $e)` and `formatExceptionData(Throwable $e)`, which works correctly in the scenarios I've tested but obviously isn't particularly clean as the code refers to Exceptions everywhere.

jamiehd:
Any update on this? Is there any plans to support php7 or should this project be considered abandoned? If I work on a PR will it be accepted?

Thanks

marceux:
@jamiehd I think pull requests are monitored and the code is updated occasionally; please make the change because I'd need it, hahaha

barryvdh:
Yes, I think you could probably remove the type hint and just add phpdocs, right?
I will accept PR's :)

JCMais:
Any plans for a release with support for PHP 7?

DevNIX:
Following the conversation in #279:

I feel like adding an error with the `addException` method, as the message will appear in the "Exceptions" tab on the debug bar, besides the `addThrowable` should be the best idea.

Maybe removing the PHP typehint and checking the instance in runtime to maintain compatibility with PHP 5 and support the `Throwable` interface if defined would be a good idea?

PR #276, opened 2016-06-09
Added php7 throwable compatability #264

jamiehd:
I've put an exception phpdoc hint in there too so PHP5 users can still make use of it.

### DIFF
--- a/src/DebugBar/DataCollector/ExceptionsCollector.php
+++ b/src/DebugBar/DataCollector/ExceptionsCollector.php
@@ -10,8 +10,6 @@
 
 namespace DebugBar\DataCollector;
 
-use Exception;
-
 /**
  * Collects info about exceptions
  */
@@ -23,9 +21,9 @@ class ExceptionsCollector extends DataCollector implements Renderable
     /**
      * Adds an exception to be profiled in the debug bar
      *
-     * @param Exception $e
+     * @param \Exception|\Throwable $e
      */
-    public function addException(Exception $e)
+    public function addException($e)
     {
         $this->exceptions[] = $e;
         if ($this->chainExceptions && $previous = $e->getPrevious()) {
@@ -46,7 +44,7 @@ class ExceptionsCollector extends DataCollector implements Renderable
     /**
      * Returns the list of exceptions being profiled
      *
-     * @return array[Exception]
+     * @return array[\Exception|\Throwable]
      */
     public function getExceptions()
     {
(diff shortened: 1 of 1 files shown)


barryvdh:
Hmm, this is good but the problem now is that we will break BC, so this would warrant a new release for just this.. See https://3v4l.org/bUWPp

An alternative would be the create a new collector, but that aint pretty either.
(So deprecate the ExceptionsCollector, create a new ThrowableCollector)

GrahamCampbell:
This shoulodn't be needed?

GrahamCampbell:
Just transform it like laravel does.

barryvdh:
what does Laravel do? Convert to plain exception, with throwable as previous?

GrahamCampbell:
Uses the symfony stuff.

GrahamCampbell:
That's why it's exception handler has kept the Exception typehint.

GrahamCampbell:
https://github.com/laravel/framework/blob/5.2/src/Illuminate/View/Engines/PhpEngine.php#L41-L47

barryvdh:
Okay yeah, so that would need to happen in application implementing debugbar, not much I can do about that. (Except for laravel-debugbar of course)

jamiehd:
Well that works for me - I'm only using it in Laravel anyway. People who are using it directly can come up with their own solutions for now I suoppose. Will probably be worth releasing this as a new version in the future for people who are using their own implementations though.

GrahamCampbell:
> so that would need to happen in application implementing debugbar

yes

jamiehd:
I've had a look into this and it seems that due to the nested exceptions in getPrevious() this is a bit harder to implement. It's not enough to just convert the main exception - it needs to somehow go through all of the previous exceptions and check for anything that doesn't implement Exception, but I can't work out a way of doing this in the Laravel package without recreating the whole chain of exceptions.

I found I could get it to work by changing both LaravelDebugbar@modifyResponse and ExceptionsCollector@addException, but that would require both packages to be updated and also makes the symfony bits a dependency.

_LaravelDebugbar@modifyResponse_

        if (isset($response->exception)) {

            if($response->exception instanceof Exception) {

                $this->addException($response->exception);

            } else {

                $this->addException(new FatalThrowableError($response->exception));

            }

        }



_ExceptionsCollector@addException_


            if($previous instanceof Exception) {

                $this->addException($previous);

            } else {

                $this->addException(new FatalThrowableError($previous));

            }

        }

If someone can suggest a different way for me to approach this I'm happy to spend a bit of time playing around.

GrahamCampbell:
> I've had a look into this and it seems that due to the nested exceptions in getPrevious() this is a bit harder to implement.

This was already fixed in symfony. I fixed it myself actually. What version are you using?

jamiehd:
Well I was only creating a new FatalThrowableError if $response->exception wasn't inherited from Exception, which meant that if the top level exception was ok, but one of the previous wasn't then it would throw an error.

I've just tried removing the check and having it create a FatalThrowableError every time and this does fix it, but it doesn't feel right to be using it in cases where a standard exception is thrown. Do you think it's acceptable as a fix to just create it every time or should I check if any of the previous are not an exception?

PR #277, opened 2016-06-10

jamiehd:


src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -29,6 +31,9 @@ public function addException(Exception $e)
     {
         $this->exceptions[] = $e;
         if ($this->chainExceptions && $previous = $e->getPrevious()) {
+            if($previous instanceof Error) {
```
GrahamCampbell: Better to say not instance of exception

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -88,17 +93,11 @@ public function formatExceptionData(Exception $e)
         );
     }
 
-    /**
-     * @return string
```
GrahamCampbell: ?

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -88,17 +93,11 @@ public function formatExceptionData(Exception $e)
...
-     */
     public function getName()
     {
         return 'exceptions';
     }
 
-    /**
-     * @return array
```
GrahamCampbell: ?

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -10,7 +10,9 @@
 
 namespace DebugBar\DataCollector;
 
+use Error;
 use Exception;
+use Symfony\Component\Debug\Exception\FatalThrowableError;
```
reyostallenberg: Where is the requirement for this?

src/DebugBar/DataCollector/ExceptionsCollector.php
jamiehd: Welp, didn't even see that. I'll get it sorted now. I blame my crazy-efficient IDE.

src/DebugBar/DataCollector/ExceptionsCollector.php
jamiehd: Good spot. Thanks

src/DebugBar/DataCollector/ExceptionsCollector.php
barryvdh: Why? Should have same effect right?

PR #278, opened 2016-06-10

jamiehd:
Fixed the issues with the last PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "require": {
         "php": ">=5.3.0",
         "psr/log": "^1.0",
-        "symfony/var-dumper": "^2.6|^3.0"
+        "symfony/var-dumper": "^2.6|^3.0",
+        "symfony/debug": "2.8.*|3.0.*"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0|^5.0"
--- a/src/DebugBar/DataCollector/ExceptionsCollector.php
+++ b/src/DebugBar/DataCollector/ExceptionsCollector.php
@@ -10,7 +10,9 @@
 
 namespace DebugBar\DataCollector;
 
+use Error;
 use Exception;
+use Symfony\Component\Debug\Exception\FatalThrowableError;
 
 /**
  * Collects info about exceptions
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -19,7 +19,8 @@
     "require": {
         "php": ">=5.3.0",
         "psr/log": "^1.0",
-        "symfony/var-dumper": "^2.6|^3.0"
+        "symfony/var-dumper": "^2.6|^3.0",
+        "symfony/debug": "2.8.*|3.0.*"
```
reyostallenberg: So if symfony/debug 3.1 is released you don't want it to upgrade?
I think it should be `"symfony/debug": "^2.8|^3.0"`
Take a look at [Composers documentation on version numbers](https://getcomposer.org/doc/articles/versions.md)

If you had created unit tests you didn't import the `Symfony\Component\Debug\Exception\FatalThrowableError` in `DebugBar\DataCollector\ExceptionsCollector` in the first place without the requirement of symfony/debug

gtjenkins: I tested it using a laravel install, so I imported that version of symfony/debug because it was the one defined in laravel's composer.json and I knew it worked. This is why I forgot to add to composer.json initially - it was already a requirement of laravel so I didn't need to add it again.

jamiehd: I tested it using a laravel install, so I imported that version of symfony/debug because it was the one defined in laravel's composer.json and I knew it worked. This is why I forgot to add to composer.json initially - it was already a requirement of laravel so I didn't need to add it again.

barryvdh: Yes please be as permissive as possible. If 2.6 is possible, use that as minimum. Also future symfony versions for 3.x should be compatible.

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -29,6 +31,9 @@ public function addException(Exception $e)
     {
         $this->exceptions[] = $e;
         if ($this->chainExceptions && $previous = $e->getPrevious()) {
+            if($previous instanceof Error) {
```
barryvdh: Cs: space between if and parenthese.

GrahamCampbell:
You know you don't need to close the PR?

jamiehd:
No I did not. I have never submitted a PR before.

barryvdh:
You can just add extra commits to your branch, even after the PR is submitted. I will merge those commits together when I accept it.

jamiehd:
Ah I didn't realise you could squash them on the merge. That would've saved a lot of effort. Thanks for the tip.

PR #279, opened 2016-06-11

jamiehd:


composer.json
```diff
@@ -19,7 +19,8 @@
     "require": {
         "php": ">=5.3.0",
         "psr/log": "^1.0",
-        "symfony/var-dumper": "^2.6|^3.0"
+        "symfony/var-dumper": "^2.6|^3.0",
+        "symfony/debug": "^2.6|^3.0"
```
GrahamCampbell: should be sorted alphabetically

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -29,6 +31,9 @@ public function addException(Exception $e)
     {
         $this->exceptions[] = $e;
         if ($this->chainExceptions && $previous = $e->getPrevious()) {
+            if ($previous instanceof Error) {
```
GrahamCampbell: `!$previous instanceof Exception` would be more correct

barryvdh: I disagree, I think this is more readable and more explicitly states the purpose of the code.

barryvdh: And while https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Debug/Exception/FatalThrowableError.php accepts Throwable, it seems to be explicitly aimed at the Error class.

barryvdh:
Thanks!

GrahamCampbell:
@barryvdh I don't agree. Throwables don't have to extend Error, and at no point does symfony every typehint `Error`, for that exact reason.

GrahamCampbell:
Something can be throwable, but not be an error or an exception.

barryvdh:
What then?

GrahamCampbell:
I was thinking about forwards compatibility. Currently, PHP has hacked this to bits so that we'd end up with:

```
Fatal error: Class Foo cannot implement interface Throwable, extend Exception or Error instead
```

barryvdh:
Hmm, fair enough. Submit a PR and I'll accept.

barryvdh:
An alternative would be to add `addThrowable` method, because there internally isn't a reason we couldn't just use the Throwables directly. We just can't change the addException typehint.

PR #281, opened 2016-06-11
Add Throwables, deprecate exceptions

barryvdh:
Instead of transforming them to Exceptions, we could still just parse them.
We need to deprecate the public methods, because of the typhints, so implementations can use the new methods instead.
cc @jamiehd @GrahamCampbell

### DIFF
--- a/src/DebugBar/DataCollector/ExceptionsCollector.php
+++ b/src/DebugBar/DataCollector/ExceptionsCollector.php
@@ -25,15 +25,23 @@ class ExceptionsCollector extends DataCollector implements Renderable
      * Adds an exception to be profiled in the debug bar
      *
      * @param Exception $e
+     * @deprecated in favor on addThrowable
      */
-    public function addException(Exception $e)
+    public function addException($e)
+    {
+        $this->addThrowable($e);
+    }
+
+    /**
+     * Adds a Throwable to be profiled in the debug bar
+     *
+     * @param \Throwable|\Exception $e
+     */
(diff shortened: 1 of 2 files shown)


src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -25,15 +25,23 @@ class ExceptionsCollector extends DataCollector implements Renderable
      * Adds an exception to be profiled in the debug bar
      *
      * @param Exception $e
+     * @deprecated in favor on addThrowable
      */
-    public function addException(Exception $e)
+    public function addException($e)
```
GrahamCampbell: probs best to keep the typehint

src/DebugBar/DataCollector/ExceptionsCollector.php
```diff
@@ -25,15 +25,23 @@ class ExceptionsCollector extends DataCollector implements Renderable
...
+    {
+        $this->addThrowable($e);
+    }
+
+    /**
+     * Adds a Throwable to be profiled in the debug bar
+     *
+     * @param \Throwable|\Exception $e
```
GrahamCampbell: no need for the duel phpdoc

src/DebugBar/DataCollector/ExceptionsCollector.php
barryvdh: Oh oops, that was the entire point indeed :p

PR #282, opened 2016-06-16
Show caller in Propel tab

gisostallenberg:
Shows something like:
image
In the Propel tab

### DIFF
--- a/src/DebugBar/Bridge/PropelCollector.php
+++ b/src/DebugBar/Bridge/PropelCollector.php
@@ -202,11 +202,81 @@ class PropelCollector extends DataCollector implements BasicLogger, Renderable, 
             'duration' => $duration,
             'duration_str' => $this->formatDuration($duration),
             'memory' => $memory,
-            'memory_str' => $this->formatBytes($memory)
+            'memory_str' => $this->formatBytes($memory),
+            'caller' => $this->getCaller(),
         );
         $this->accumulatedTime += $duration;
         $this->peakMemory = max($this->peakMemory, $memory);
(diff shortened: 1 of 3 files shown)


barryvdh:
Probably good to standardise this in the QueryCollector. I'm doing something similar here: https://github.com/barryvdh/laravel-debugbar/blob/7668d62768cbd6b9f9ab8eebb11d6647869d5195/src/DataCollector/QueryCollector.php#L200-L232
But there I'm misusing the existing `stmt_id` key.

barryvdh:
This is what it looks like for me: 
Example

gisostallenberg:
@barryvdh QueryCollector is not in this repo. Should I add for instance ORMCollector and extend DoctrineCollector, Propel2Collector and PropelCollector from it? I have no easy means of checking DoctrineCollector and Propel2Collector though.
Skipping the various items in the backtrace is very specific to Propel.

gisostallenberg:
@barryvdh can you answer the questions above? I think this addition is very valuable, so I hope we can craft it in a way that suits both you and me.

barryvdh:
I think we should probaby create a AbstractQueryCollector with some useful methods, and extend from that.

gisostallenberg:
Ok, I will add an abstract layer like that, but I'm not able to generate proper implementations for DoctrineCollector, Propel2Collector which I will default to 'first call not in the debugbar trace'

gisostallenberg:
@barryvdh can you confirm that when I will add a change for Propel you will be able to add DoctrineCollector or Propel2Collector?

PR #283, opened 2016-06-20
renderhead by version

myphper:
renderhead by version, /debugbar/debugbar.js?v=1.0.0.1

### DIFF
--- a/src/DebugBar/JavascriptRenderer.php
+++ b/src/DebugBar/JavascriptRenderer.php
@@ -768,18 +768,21 @@ class JavascriptRenderer
      *
      * Only useful if Assetic is not used
      *
+     * @param string $v current version of static resource
      * @return string
      */
-    public function renderHead()
+    public function renderHead($v = null)
     {
         list($cssFiles, $jsFiles) = $this->getAssets(null, self::RELATIVE_URL);
         $html = '';
 
         foreach ($cssFiles as $file) {
+            $file = $v != null ? $file .'?v='. $v : $file;
             $html .= sprintf('<link rel="stylesheet" type="text/css" href="%s">' . "\n", $file);
         }
 
         foreach ($jsFiles as $file) {
+            $file = $v !=null ? $file .'?v='. $v : $file;
             $html .= sprintf('<script type="text/javascript" src="%s"></script>' . "\n", $file);
         }
 


barryvdh:
What about using the hash of the changed files, instead of the version? Because the assets change when adding some collectors with different widgets or configuration.

Something like this: https://github.com/magento-hackathon/magento2-toolbar/blob/6ad03dd3367ed5e39102a7ae7d796222b0955973/JavascriptRenderer.php#L21-L50

barryvdh:
So something like this:

```php
	 /**
     * Renders the html to include needed assets
     *
     * Only useful if Assetic is not used
     *
     * @param string $v current version of static resource
     * @return string
     */
    public function renderHead()
    {
        list($cssFiles, $jsFiles) = $this->getAssets(null, self::RELATIVE_URL);
        $html = '';
        foreach ($cssFiles as $file) {
            $file = $file . '?v=' . $this->getAssetsHash('css');
            $html .= sprintf('<link rel="stylesheet" type="text/css" href="%s">' . "\n", $file);
        }
        foreach ($jsFiles as $file) {
            $file = $file . '?v=' . $this->getAssetsHash('js');
            $html .= sprintf('<script type="text/javascript" src="%s"></script>' . "\n", $file);
        }
        if ($this->enableJqueryNoConflict && !$this->useRequireJs) {
            $html .= '<script type="text/javascript">jQuery.noConflict(true);</script>' . "\n";
        }
        return $html;
    }
	
    /**
     * Get the hash of the included assets, based on filename and modified time.
     *
     * @param string $type 'js' or 'css'
     * @return string
     */
    protected function getAssetsHash($type)
    {
        $assets = [];
        foreach ($this->getAssets($type) as $file) {
            $assets[$file] = filemtime($file);
        }
        return md5(serialize($assets));
    }
```

myphper:
well,good idea.but sometimes we must control the static resource version,at least when i use CDN

barryvdh:
How are you using a CDN?

barryvdh:
If you are using a CDN, you specifiy the version in the URL right? Cause otherwise you will get out-of-sync versions.

myphper:
yes

myphper:
thx ,your solution is ok if static resource file and php file run on the same server , but now i didn't

PR #287, opened 2016-07-15
Aggregate Timeline Measures

dan-iway:
This aggregates all time measures with the same label and displays the count and total duration in a table on the Timeline tab. 

image

I am not sure this will be useful for others but it has helped me a lot to debug performance issues when there are many small steps involved and it is not clear which of the small steps are adding up most. 
Without this code would have to be rearranged to fit into a single measurement.


spotman:
:+1: for merging that! Very useful feature.

PR #292, opened 2016-08-25
Added Grav to integrations

rhukster:
I've added Grav to the list of integrations as we've been using DebugBar in Grav for the past couple of years.  Thanks for the great project!


barryvdh:
Is it integrated in Grav itself? Otherwise better link to the integration.

rhukster:
Yes it's built into Grav itself. https://learn.getgrav.org/advanced/debugging#debug-bar

rhukster:
Cheers!

PR #297, opened 2016-10-12
Added array reverse to find results

phpcodemonkey:
The OpenHandler modal returns results in oldest first order by default, changed storage find method so newest entries are at the top.


james-johnston-thumbtack:
This is also something that has annoyed us; could this be merged? @barryvdh

PR #303, opened 2016-11-22
Add Xdebug Link functionality

amenk:
https://github.com/barryvdh/laravel-debugbar/issues/566


barryvdh:
So should we add this to more places? Eg. exceptions and possibly backtraces from queries etc?

amenk:
Yes, that was my intention :) Can you do this or shall I? not so familiar with the Debugbar Code - maybe you are faster

zschuessler:
For an easier user experience, you can enact the call over AJAX. That way the user isn't redirected to a localhost api page. (At least with my setup for PhpStorm, clicking a link redirects me to a localhost page)

I have a hacky workaround that demonstrates a solution:

```javascript
 if (typeof tpl.xdebug_link !== 'undefined') {
   $('<a class="xdebug-link" href="' + tpl.xdebug_link + '">VIEW FILE</a>').addClass(csscls('editor-link')).appendTo(li);

  // Load xdebug link over AJAX GET request
   $(document).on('click', '.xdebug-link', function(e) {
     e.preventDefault();
     var xdebugLink = e.target.href;
     $.get(xdebugLink.replace('http', 'https'));
   })
 }
```

Note that I'm serving my local content over ssl, so I also added a line to force a protocol. The only relevant piece is the fact the link is loaded by an AJAX GET request.

Voila, no more user redirection when clicking the links.

PR #320, opened 2017-02-18
Use system fonts

sebdesign:
Use system fonts to match current OS UI chrome for `sans-serif` and
`monospace` fonts, giving a more consistent look with the OS and other
developer tools in the browser.

As as side note, this technique is also used by GitHub.
Also, font-awesome icons on the same line with monospace text are now correctly aligned.

Read more: https://css-tricks.com/snippets/css/system-font-stack/

Here is a screenshot on macOS Sierra:

laravel-debugbar-queries


barryvdh:
Thanks!

Issue #328, opened 2017-05-02
PHP 7.1 | "A non well formed numeric value encountered"


voku:
in "composer/maximebf/debugbar/src/DebugBar/DataCollector/TimeDataCollector.php"

```php
     * Returns the duration of a request
     *
     * @return float
     */
    public function getRequestDuration()
    {
        if ($this->requestEndTime !== null) {
            return $this->requestEndTime - $this->requestStartTime;
        }
        return microtime(true) - $this->requestStartTime;
        ...
```

object(DebugBar\DataCollector\TimeDataCollector)#1675 (5) { ["requestStartTime":protected]=> string(15) "1493739115,4072" ["requestEndTime":protected]=> float(1493739115,6443) ["startedMeasures":protected]=> array(0) { } ["measures":protected]=> array(0) { } ["dataFormater":protected]=> NULL }

--> fix

```php
return (float)$this->requestEndTime - (float)$this->requestStartTime;
```

https://bugs.php.net/bug.php?id=73468 (Status: Not a bug)

barryvdh:
Can you submit a PR?

caugner:
@barryvdh This issue could be closed.

caugner:
@barryvdh Would you mind closing this issue (which has been fixed by #329)?

PR #329, opened 2017-05-04
Fix PHP 7.1 | "A non well formed numeric value encountered"

voku:
This change is [Reviewable](https://reviewable.io/reviews/maximebf/php-debugbar/329)


voku:
fix for #328

Issue #330, opened 2017-05-11
Error extending templates (Conditional Inheritance)


manusreload:
If you do a conditional inheritance (https://twig.sensiolabs.org/doc/2.x/tags/extends.html#conditional-inheritance) the resolved template is "".

*Template File*

```
{% extends this.response.ajax == true ? "templates/modal.twig" : "templates/base.twig" %}

{% block title %}Pendiente{% endblock %}
{% block menu %}
    {% include "components/menu_taller.twig" %}
{% endblock %}
{% block body %}
...
```

*Exception throws*

```
Unable to find template "" (looked into: ././modules/glplanning/view, ././modules/glplanning/view/includes, ./view, ././vendor/gestionlan/framework/modules/installer/view, ././vendor/gestionlan/framework/modules/incidencias/view, ././vendor/gestionlan/framework/modules/api/views, ././vendor/gestionlan/framework/modules/admin/views, ././modules/gti/view, ././modules/notas/view, ././modules/vsustitucion/view, ././modules/vacaciones/view, /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/gestionlan, /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/gestionlan/framework, /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/gestionlan/framework/views, /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/gestionlan/framework/modules) in "/taller/pendiente.twig" at line 1.</h3> at /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/twig/twig/lib/Twig/Loader/Filesystem.php:232<pre>#0 /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/twig/twig/lib/Twig/Loader/Filesystem.php(150): Twig_Loader_Filesystem->findTemplate(Object(DebugBar\Bridge\Twig\TraceableTwigTemplate))
#1 /Users/manus/Dropbox/Trabajo/GestionLan/GLAutoBox/vendor/twig/twig/lib/Twig/Environment.php(329): Twig_Loader_Filesystem->getCacheKey(Object(DebugBar\Bridge\Twig\TraceableTwigTemplate))
[...]

PR #331, opened 2017-05-20
Fix twig template instance check

gsouf:
Fix for  #330

Twig was unable to check for template being instance of self because Traceable twig template was not extending it.
Please note that since version 2 of twig, Twig_Template does not implement Twig_TemplateInterface  and Twig_TemplateInterface will be removed in version 3.x and thus phpdebugbar should do so.


gsouf:
ping @maximebf

gsouf:
Thank you @barryvdh

PR #333, opened 2017-07-13
localize hljs's css to phpdebugbar

HandsomeMatt:
currently these styles will override anything in my application, localizing them to only the debugbar seems like a good solution.


barryvdh:
This is good, but difficult to maintain with future updates.

PR #355, opened 2017-09-01
Fix for memory overrun in RedisStorage

ilnicki:
Patch increases number of possible count of records loaded from Redis DB by picking them one by one instead of bulk retrieving. HGETALL causes FatalErrorException with "Allowed memory size of 134217728 bytes exhausted" on around of 2500 records, Patch leads to a little speedup even on small amount or records (about 3-4%) and also allows to work on 1500+ records as well.
Problem is described in barryvdh/laravel-debugbar#417

### DIFF
--- a/src/DebugBar/Storage/RedisStorage.php
+++ b/src/DebugBar/Storage/RedisStorage.php
@@ -51,8 +51,8 @@ class RedisStorage implements StorageInterface
     public function find(array $filters = array(), $max = 20, $offset = 0)
     {
         $results = array();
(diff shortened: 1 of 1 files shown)


barryvdh:
Why close?

ilnicki:
Want to test with phpredis. Currently tested only with phredis. Also have an idea to use hscan method. It can be better solution. Storage provider is working well with this patch, but hardly needs a full rework because it's barely useful afler couple hours of site testing. Storage reaches 200-300 mb and every list opening makes a 10-15 seconds delay. I'll try to find a simple and beautiful solution for the problem.

PR #357, opened 2017-09-07
RedisStorage performance fix

ilnicki:
Patch increases possible number of records stored and loaded from Redis DB by separating metadata and collector's data and storing them in different hashes. Retrieving is done with HSCAN instead of HGETALL, which causes FatalErrorException with "Allowed memory size of 134217728 bytes exhausted" on around of 2500 records. HSCAN takes records by portions what leads to a little speedup even on small amount or records (about 3-4%) and also allows to work on 7500+ records as well.
Tested for 5 days and still goes well. Current version takes about 10sec to load a request list on ~4500 records set, while this does the same in 0.8sec on 7700+.
Problem is described in barryvdh/laravel-debugbar#417

### DIFF
--- a/src/DebugBar/Storage/RedisStorage.php
+++ b/src/DebugBar/Storage/RedisStorage.php
@@ -34,7 +34,9 @@ class RedisStorage implements StorageInterface
      */
     public function save($id, $data)
     {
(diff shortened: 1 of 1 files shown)


ilnicki:
14000+ records and still loaded in about 2sec.

Issue #324, opened 2017-03-12
Twig 2 and the default collector


hultberg:
While adding the default `TwigCollector` with an `TraceableTwigEnvironment` I get this error:
```
Declaration of DebugBar\Bridge\Twig\TraceableTwigEnvironment::setLexer(Twig_LexerInterface $lexer) should be compatible with Twig_Environment::setLexer(Twig_Lexer $lexer)
```
The `LexerInterface` has been [deprecated in twig since v1.12](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/LexerInterface.php#L17) and has been removed in v2.0. I suggest changing the declaration of `TraceableTwigEnvironment::setLexer(Twig_LexerInterface)` to be `TraceableTwigEnvironment::setLexer(Twig_Lexer)`. However, this might break some projects if they have implemented `Twig_LexerInterface` and not extending `Twig_Lexer`.

While browsing the file `TraceableTwigEnvironment`, it seems it should be reviewed to work with twig 2.0.

Versions:
```
PHP: 7.1
Twig: 2.2.0
Debugbar: 1.13.1
```

Issue #243, opened 2015-12-24
Twig : 0 templates were rendered


Merrick28:
Hello everyone,

I just installed this tool, and it works great, except for Twig. 
I have to precise that I don't use Symfony, it's a non-framework project.

I think I've made a mistake somewehere.

Here is my (simplified) code :
```
use DebugBar\StandardDebugBar;
$debugbar         = new StandardDebugBar();

$loader = new Twig_Loader_Filesystem(CHEMIN . '/templates');
$twig = new DebugBar\Bridge\Twig\TraceableTwigEnvironment(new Twig_Environment($loader, array()));

$debugbar->addCollector(new DebugBar\Bridge\Twig\TwigCollector($twig));
$optionsTwig = array(
   'TITLE' => 'Test de titre',
);
echo $twig->render('page.twig',$optionsTwig);
echo $debugbarRenderer->render();
```
I can see the phpdebugbar, and there is a twig tab, but it's always empty. What did I miss ?

barryvdh:
The Twig loader is currently a bit out-dated. It should be updated but don't have the time yet.

Not really sure what's wrong here.

Merrick28:
Ok, it's not a big deal for me, Twig isn't the topic I would like to monitor :)
Thanks for your great work.

PR #362, opened 2017-10-06
Feature/twig profiler

### LINKED ISSUE #192: Twig Profiler
There is a WIP PR for Twig to add a Profiler extension: https://github.com/twigphp/Twig/pull/1597

The linked PR for Symfony Webtoolbar shows how this could be implemented. Probably use that instead of the current TwigCollector when it's available. We could also use the TimeData collector to build the call graph.

Keeping this open as a reference, untill the PR is merged. Then we could do an initial PR.

timri:
This adds a Debugbar\Bridge\TwigProfileCollector utilizing the class Twig_Extension_Profiler (provided by twig) as an alternative to (replacement for) Debugbar\Bridge\Twig\TwigCollector.

The first commit contains the base feature, the second the documentation and the third contains a derived Twig_Extension to also measure the rendering on the TImeline-Tab.

This should fix #192 and probably also #243 and #324

### DIFF
--- a/docs/bridge_collectors.md
+++ b/docs/bridge_collectors.md
@@ -113,3 +113,12 @@ You need to inject the root-`Twig_Profiler_Profile` into the collector:
     $env->addExtension(new Twig_Extension_Profiler($profile));
     $debugbar->addCollector(new DebugBar\Bridge\TwigProfileCollector($profile));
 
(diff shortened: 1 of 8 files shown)


yanikore:
When using namespace and extending blocks, the profiler is called multiple times and the measures are lost. `TimeableTwigExtensionProfile->leave()` crashes because it will try to stop a measure that "has not been started yet".

PR #370, opened 2017-11-18
Support for Symfony 4

Nyholm:
#SymfonyConHackday2017

PR #373, opened 2017-12-05
allow symfony 4

kornrunner:
Ease constraint on `symfony/var-dumper` to allow Symfony 4.

If this gets merged in, could it also be tagged as a patch? Thank you!


kornrunner:
Duplicate of #370

PR #393, opened 2018-09-18
Stop TracedStatement throwing warnings for objects

AnthonyPorthouse:
The current behaviour of `TracedStatement::checkParameters()` will cause warnings to be raised when objects are passed in as a parameter. This should not always be the case.

This PR adds support for both objects with and without `__toString()` methods.

If an object has a `__toString()` method, then we will call that and use that as our parameter.

If it does not, then we will short circuit the encoding check and just make it show up as `[BINARY DATA]`.

Another option, which is preferred would be to add another possible value such as `[OBJECT classname]` which would differentiate it from standard binary data.


barryvdh:
I think this is already fixed.

PR #398, opened 2018-11-12
Add body message to mail

Jonathanm10:
I'd like to have the body but also to be able to interpret HTML. To have a better look at what the user will received.
Here's an example
image

Let me know if you think this should be done in an other way.

Cheers,

Jonathan


jeanmonod:
This is a really nice feature! Can we have it merged? please 🙏

barryvdh:
This will probably make it a lot slower to render. And I don't know if any styling etc would be able to leak through to the webpage.
It should be optional at least.

angeljqv:
> any styling etc would be able to leak through to the webpage

maybe using an iframe to avoid css collisions

barryvdh:
I think it would be cool to open the body as a popup?

parallels999:
Fixed on #617

PR #404, opened 2019-01-13
Improving some PHPDoc types and array notation for readability

igorsantos07:
I found some parts of TraceablePDO-related code a bit confusing (using it for extra tasks on my project), so I'm clearing it up here :)

I would also like to ask if it's needed to keep all those repeated methods inside TraceablePDO; if `__call()` already forwards all calls, why pollute the new class file with meaningless forwarding methods?


barryvdh:
Thanks!

PR #407, opened 2019-01-21
Makes the PDO wrapper transparent

igorsantos07:
It's easy to auto-wrap the PDO object inside the collector. As the wrapped object is only useful to the collector, it seems this responsibility should be internal.

### DIFF
--- a/src/DebugBar/DataCollector/PDO/PDOCollector.php
+++ b/src/DebugBar/DataCollector/PDO/PDOCollector.php
@@ -24,7 +24,7 @@ class PDOCollector extends DataCollector implements Renderable, AssetProvider
      * @param TraceablePDO $pdo
      * @param TimeDataCollector $timeCollector
      */
-    public function __construct(TraceablePDO $pdo = null, TimeDataCollector $timeCollector = null)
+    public function __construct(\PDO $pdo = null, TimeDataCollector $timeCollector = null)
     {
         $this->timeCollector = $timeCollector;
         if ($pdo !== null) {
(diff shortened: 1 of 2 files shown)


barryvdh:
Do the queries still get logged this way?

igorsantos07:
Yep! We've been using it this fork on our company since I created it, and re-using the TraceablePDO for extra operations.

It maintains the same functionality while reducing the user friction - they don't need to know there's a PDO wrapper :)

igorsantos07:
Ping @barryvdh / @maximebf :)

src/DebugBar/DataCollector/PDO/PDOCollector.php
```diff
@@ -65,11 +65,14 @@ public function getSqlQuotationChar()
...
      * @param string $name Optional connection name
      */
-    public function addConnection(TraceablePDO $pdo, $name = null)
+    public function addConnection(\PDO $pdo, $name = null)
     {
         if ($name === null) {
             $name = spl_object_hash($pdo);
         }
+        if ($pdo instanceof \PDO) {
+            $pdo = new TraceablePDO($pdo);
```
barryvdh: Isn't this always true? Shouldn't it check if it isn't an instanceof TraceablePDO?

igorsantos07: :man_facepalming:

igorsantos07: I got caught on this one right now too. But actually `PDOCollector::__construct()` accepts `null` as the `$pdo` param, thus this check is needed so it doesn't try to envelope null instead of a PDO instance in those cases.

igorsantos07: I'm not sure if it's an unsent or missed notification, or simply the maintainers busy with life and stuff, but anyway I'll ping @barryvdh again. Sorry if that's spammy, but this is fixed :sweat_smile:

igorsantos07:
This was resolved @barryvdh

PR #409, opened 2019-03-04
JavaScriptRenderer: support deleting stacked data

caugner:
This pull request adds `$deleteStackedData`, a parameter similar to `$renderStackedData` which automatically deletes stacked data alongside rendering it.

This change allows the use case where stacked data should only be rendered once.

A workaround is to delete the data manually after rendering:

```php
$html = $bar->getRenderer()->render();
if ($bar->hasStackedData()) {
  $bar->getStackedData(true);
}
echo $html;
```


caugner:
Nevermind, it already deletes it automatically.

PR #411, opened 2019-03-18
FIX PHP 5.4 compat for using function return value in write context

robbieaverill:
This restores compatibility with older PHP versions. [v1.15.0 still allows PHP 5.3 or greater](https://github.com/maximebf/php-debugbar/blob/v1.15.0/composer.json#L20).

It'd be great if you could release this in a 1.15.x patch, I realise it'll be slightly annoying since other commits exist in master already which bump these requirements.

You can see the error here: https://travis-ci.org/lekoala/silverstripe-debugbar/jobs/508134678#L528

### DIFF
--- a/src/DebugBar/DataCollector/DataCollector.php
+++ b/src/DebugBar/DataCollector/DataCollector.php
@@ -175,8 +175,9 @@ abstract class DataCollector implements DataCollectorInterface
      */
     public function getXdebugLinkTemplate()
     {
-        if (empty($this->xdebugLinkTemplate) && !empty(ini_get('xdebug.file_link_format'))) {
-            $this->xdebugLinkTemplate = ini_get('xdebug.file_link_format');
+        $fileLinkFormat = ini_get('xdebug.file_link_format');
+        if (empty($this->xdebugLinkTemplate) && !empty($fileLinkFormat)) {
+            $this->xdebugLinkTemplate = $fileLinkFormat;
         }
 
         return $this->xdebugLinkTemplate;
--- a/src/DebugBar/DataCollector/DataCollector.php
+++ b/src/DebugBar/DataCollector/DataCollector.php
@@ -91,13 +91,13 @@ abstract class DataCollector implements DataCollectorInterface
             $file = strtr($file, $this->xdebugReplacements);
         }
 
-        $url = strtr($this->getXdebugLinkTemplate(), ['%f' => $file, '%l' => $line]);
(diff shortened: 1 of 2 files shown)


barryvdh:
So I would need a separate branch for 1.15 and then merge this?

robbieaverill:
Yeah that’s be great if you wouldn’t mind. I’ll rebase the PR afterwards

leomoty:
From my experience this is not entirely true for php 5.3.3 (or lower), which is the case I still have to handle in a legacy system.

the array `[ ]` syntax is only supported in php 5.4 and it is used in the same file you changed:
https://github.com/leomoty/php-debugbar/commit/bb7d7e521d81f0747ba88e05021fdbef1eb7f7dc

`$fileinfo->getExtension()` is only available after php 5.3.6:

https://github.com/leomoty/php-debugbar/commit/51df756c566a7a7de5e58bb986fce498d1aa566d

Anyways, my point is, either bump composer.json support for `>5.4.0` or you are likely to need my extra fixes.

robbieaverill:
It seems strange to remove support for PHP 5.4 in a patch release (to me). Perhaps either someone using PHP 5.3 needs to whip over and fix the incompatible syntax, or revert any commits that broke PHP 5.3 compatibility. In my case I'm also (very slightly) concerned with PHP 5.3 as well as PHP 5.4.

leomoty:
If you check the commits I linked, I fixed the `array square brackets`, `getExtension` and the `ini_get`. I didn't expect this to be accepted upstream, so that is why I went that route.

And I agree that it seems pretty weird, I just meant that the repository in that specific tag wasn't really PHP 5.3.0 compliant.

robbieaverill:
@leomoty maybe you could create a pull request from your commits in order to restore compatibility for it =)

leomoty:
@robbieaverill or you can just add my commits to your PR? either way, we need to wait for the 1.15 branch, right?

robbieaverill:
Thanks @leomoty, I've cherry picked your commits into this pull request

robbieaverill:
@barryvdh anything I can do to get this merged in?

robbieaverill:
Hi @barryvdh, not trying to be annoying here, but would you be able to merge this? If you'd like extra maintenance help, I'm happy to get involved as well.

robbieaverill:
@barryvdh feel free to reopen this PR if you'd like to look at it in the future, for now we'll drop support for PHP 5.3 and 5.4 instead. Thanks.

leomoty:
I don't have that option, but i'll just keep using my fork till we are finally able to get rid of PHP 5.3

PR #417, opened 2019-05-30
Changing single click to double click when showing/hiding details.

FallenBG:
This change will enable users to actually be able to select and copy the messages/errors/exceptions/params that are expanded. Right now trying to select anything will collapse the block preventing the selection.


barryvdh:
What do you think about https://github.com/maximebf/php-debugbar/pull/608
Preventing click when selection is active

barryvdh:
Fixed with #608

Issue #421, opened 2019-09-23
Fetch support


stevelacey:
Any plans to support `fetch` in addition to `xhr`?

I don't see a way to hook php-debugbar up to Apollo currently, whilst it works just fine with Axios, is lack of fetch support the reason? I came across the docs for manual integration but it seems to be `xhr` specific and throwing `fetch` as it doesn't go so well.

Would be great to hook this up to Apollo-based SPA's so I can see the horrendous number of database queries being ran easier 🙃

barryvdh:
I think this is what we do now: https://github.com/maximebf/php-debugbar/blob/2a95c20da4c6965114e5773f3aaf974e12f9c92b/src/DebugBar/Resources/debugbar.js#L1151-L1169

And this is how Symfony does it, which supports fetch: https://github.com/symfony/symfony/blob/7788612c3cb63cc44b2cabb4644a7fbd75c61e97/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Profiler/base_js.html.twig#L279-L331

So I think we should be able to support fetch also. Do you want to make a PR for that?

stevelacey:
@barryvdh yep, done, please check out #422 🤘

fjorgemota:
I think this issue can be closed, right?

parallels999:
> I think this issue can be closed, right?

Yes, must be closed

PR #422, opened 2019-09-24
Fetch support

stevelacey:
Adds support for fetch in addition to xhr (and jquery) #421

Most notably (for me anyway) this gets debugbar working in Apollo-based SPA's and GraphiQL.

### DIFF
--- a/src/DebugBar/Resources/debugbar.js
+++ b/src/DebugBar/Resources/debugbar.js
@@ -95,7 +95,7 @@ if (typeof(PhpDebugBar) == 'undefined') {
 
 
     // ------------------------------------------------------------------
-    
+
     /**
      * Base class for all elements with a visual component
      *
@@ -123,7 +123,7 @@ if (typeof(PhpDebugBar) == 'undefined') {
 
         /**
          * Called after the constructor
-         * 
+         *
          * @param {Object} options
          */
         initialize: function(options) {
@@ -137,7 +137,7 @@ if (typeof(PhpDebugBar) == 'undefined') {
 
         /**
          * Sets the value of an attribute
-         * 
+         *
          * @param {String} attr Can also be an object to set multiple attributes at once
          * @param {Object} value
          */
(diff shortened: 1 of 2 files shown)


stevelacey:
Note that due to the way we have to swap out `window.fetch` rather than the `open` method on the xmlhttprequest prototype, most users will probably need to stick `async defer` on their script(s) to track initial requests on pageload; something for docs maybe... I imagine it'd apply to a way large percent of use cases than with `xhr` given the common use-case for fetch being in clients like `Apollo` that initialize on pageload and touch `window.fetch` then rather than on the first request.

src/DebugBar/Resources/debugbar.js
```diff
@@ -1037,31 +1037,49 @@ if (typeof(PhpDebugBar) == 'undefined') {
...
-         * @param {XMLHttpRequest} xhr
+         * @param {Response|XMLHttpRequest} response
          * @return {Bool}
          */
-        handle: function(xhr) {
-             // Check if the debugbar header is available
-            if (xhr.getAllResponseHeaders().indexOf(this.headerName) === -1){
+        handle: function(response) {
+            // Check if the debugbar header is available
+            if (this.isFetch() && !response.headers.has(this.headerName + '-id')) {
```
barryvdh: Does this need a response? `this.isFetch(response)`

barryvdh:
Did you test this with bot xhr and fetch? Is this ready?

stevelacey:
@barryvdh can you tag a release 🙏

barryvdh:
I tagged it, but doesn't yet show on packagist. @maximebf can you check your Packagist integration settings?

stevelacey:
@barryvdh might be GitHub, they're going through Partially Degraded Service right now

stevelacey:
Maybe not, @maximebf help?

stevelacey:
@maximebf pint

stevelacey:
@barryvdh should note that the version constraints between the last tagged version on package and master are quite different, the minimum PHP version is bumped from 5.3 to 5.6 for one thing.

Looks like there hasn't been a tagged release in almost 2 years 😱

stevelacey:
@maximebf ping

PR #425, opened 2019-10-09
Deprecate TwigCollector and friends

### LINKED ISSUE #423: Update twig usings
current twig usings are deprecated since 2.7. for example `Twig_Environment` is now `Twig\Environment`

leomoty:
The reasoning behind this merge is to get rid of the `TwigCollector`, which seems to work only for `twig/twig <2`.

Since there is an alternative `TwigProfileCollector` which seems to work throughout v1 and v2 of Twig, I have updated the demo to use that instead.

And also cleaned the bridge documentation to reflect the now absent `TwigCollector`.

This fixes #423. WDYT @barryvdh?

### DIFF
--- a/demo/bridge/twig/index.php
+++ b/demo/bridge/twig/index.php
@@ -6,9 +6,11 @@ include __DIR__ . '/../../bootstrap.php';
 $debugbarRenderer->setBaseUrl('../../../src/DebugBar/Resources');
 
 $loader = new Twig_Loader_Filesystem('.');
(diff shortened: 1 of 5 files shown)


barryvdh:
We can't actually remove it, only deprecate it, without breaking BC.

leomoty:
oh ops, should add `@deprecated` to the classes then?

leomoty:
@barryvdh updated, PTAL.

PR #427, opened 2019-10-18
Add void return type hint for MonologCollector::write()

ekateiva:
This would fix an issue who uses `MonologCollector` with `Monolog` v2.

There is an issue when you try to use `Bridge/MonologCollector` because the abstract method signature has changed in `Monolog` v2. The diff: https://github.com/Seldaek/monolog/commit/06143b03e5c1598bfa6c6df65432010da0ecc28b#diff-d1d031ebe96084844fb53eb650bc404eR50

However, this change is a bit tricky. Because `Monolog` in v2 has increased PHP dependency to `^7.2`, also added those void return type hints. So this change could break things for people who uses `MonologCollector` with PHP version lower than 7.1.

### DIFF
--- a/src/DebugBar/Bridge/MonologCollector.php
+++ b/src/DebugBar/Bridge/MonologCollector.php
@@ -59,7 +59,7 @@ class MonologCollector extends AbstractProcessingHandler implements DataCollecto
     /**
      * @param array $record
      */
-    protected function write(array $record)
+    protected function write(array $record): void
     {
         $this->records[] = array(
             'message' => $record['formatted'],


barryvdh:
So should we make a separate v2 version?

ekateiva:
I guess we should then, in order not to break things for existing users but also allow to use this package with a new version of `monolog`. Then we should also modify PHP requirement to `"php": "^7.1",`.

Right now there is an issue that `laravel-debugbar` can not be used with Laravel v6.x and turned of `collectors.messages` setting. Sure, we can modify how it is done in `laravel-debubar` and do not use `MonologCollector ` in that case and leave everything in this package as it is still somebody will start complaining about not supported new version of Monolog. For more details see https://github.com/barryvdh/laravel-debugbar/pull/982.

barryvdh:
Yes but we can support both Monolog versions, without hardcoding a dependency. Just a MonologCollector and Monolog2Collector or something. 
In Laravel Debugbar we can then check which version is used and use the correct collector.

ekateiva:
Sounds like an approach. Would you like me to prepare this PR?

barryvdh:
Hmm, but we can make the typehint stricter when extending, so:
Monolog1 without void -> Collector with void -> works
Monolog2 with void -> Collector with void -> works
pre-7.0 -> never works

barryvdh:
So we could bump the PHP version to 7 and then merge this. People on old PHP versions just stick with the current release.

barryvdh:
This is now in v1.16-dev.

ekateiva:
@barryvdh There is also an alternative approach with `MonologCollector` and `MonologV2Collector`.
You could check the diff: https://github.com/maximebf/php-debugbar/compare/master...ekateiva:feature/support-monolog-v2?expand=1

Then in `laravel-debugbar` we could have:
```
    if ($this->hasCollector('messages')) {
        ...
    } else {
        $this->addCollector($this->getMonologCollector());
    }

    /**
     * @return AbstractMonologCollector
     * @throws Exception
     */
    private function getMonologCollector()
    {
        $logger = $this->getMonologLogger();
        if (class_exists('DebugBar\Bridge\MonologV2Collector')) {
            return new MonologV2Collector($logger);
        }

        return new MonologCollector($logger);
    }
```

barryvdh:
Yeah but I think it should work now, right? As long as it's PHP 7+

ekateiva:
^^ Yes, as long as it is PHP ^7.1, as return void types were added only in 7.1 (https://www.php.net/manual/en/migration71.new-features.php)

barryvdh:
Ah yeah, fixed

PR #436, opened 2020-01-13
Adding support to \Redis and fixing method capitalizations

igorsantos07:


barryvdh:
Does this work for both?

igorsantos07:
Never saw the question and came back here years later, now that I'm trying to get out of my own fork on the project.

Closing this since someone copied over my changes and expanded on them without proper reference. They were small anyway, at least :man_shrugging:

PR #452, opened 2020-10-06
feature for iframe propogate events out

lintaba:
this adds support for iframes to send their data to the parent frame.

also, same-origin should keep the payload safe


barryvdh:
What would this do?

lintaba:
A few projects ago we used a lot of business logic inside iframes. We needed a way to add them to the debugbar. (a bit like it happens with ajax+headers).
The debugbar is opened in the top frame, and when something happens inside the iframes, the changes are being added to the top-level debugbar.

For me it was useful, and used a lot. Just wanted to contribute it back to the community :)

PR #453, opened 2020-10-16
@inheritDoc instead of {@inheritDoc} Hacktoberfest

phpfui:
Minor issue, but for Hacktoberfest, I thought I would correct it.

{@inheritDoc} should be @inheritDoc

See [Inheritance](https://docs.phpdoc.org/latest/guides/inheritance.html)

### DIFF
--- a/src/DebugBar/DataCollector/PhpInfoCollector.php
+++ b/src/DebugBar/DataCollector/PhpInfoCollector.php
@@ -35,7 +35,7 @@ class PhpInfoCollector extends DataCollector implements Renderable
     }
 
     /**
-     * {@inheritDoc}
+     * @inheritDoc
      */
     public function getWidgets()
     {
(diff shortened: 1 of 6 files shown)


barryvdh:
In the docs it doesn't say this is correct right? Only that they are working towards it.

phpfui:
I believe they are saying @inheritDoc is the correct form for how you are using it, ie. this is documented the same way as the parent class method.  {@inheritDoc} is used to insert the inherited documentation into more documentation for the method.

I have looked around at the libraries I use, and it seems most have switched over to @inheritDoc from {@inheritDoc}.  There are a few, DeepCopy for example, that add some text, then {@inheritDoc}. Symfony seems to still use {@inheritDoc} though.

In my reading, it looks like a simple @inheritDoc is the best way to indicate this method is documented the same way as it's parent method.

I also noticed that \phpDocumentor\Reflection\DocBlock leaves {@inheritDoc} as plain text, but makes a proper tag out of @inheritDoc, which is how I found out about this.

Hope this helps clarify.

PR #467, opened 2021-03-05
Update classes to remove deprecation warnings

CColeman13:
Update to use Twig\Extension\ProfilerExtension  instead of Twig_Extension_Profiler
and \Twig\Profiler\Dumper\HtmlDumper instead of Twig_Profiler_Dumper_Html


barryvdh:
This is fixed with a new class: https://github.com/maximebf/php-debugbar/pull/479

PR #471, opened 2021-05-27
Add support for JS nonce

### LINKED ISSUE #244: Feature Request: Compatability with CSP headers
I understand its a bit of an undertaking, but I might as well ask.

Just implemented Content Security Policy headers on my project today. Turns out, inline-executed javascript is frowned upon. So, the whole building of the phpdebugbar js object in the footer of the page that adds tabs and data is troublesome. CSP guidelines would probably suggest executing the javascript in a separate file and having the data loaded asynchronously. Which I totally get is a PITA given that this data is inherently linked to how __THIS__ page performed. Some sort of session flashing might work?

It would be great if I could have CSP turned on in all environments, including the ones that we've got the debugbar running in. For now though, I'll just have to turn off CSP for developers, because we so very much love your debugbar :)

MatteoGheza:
Fixes #244


MatteoGheza:
Similar to #439

barryvdh:
This PR uses the options array, which is probably better. Not sure if we need the custom attributes though. Do you have a specific use case outside nonce?

barryvdh:
I've updated https://github.com/maximebf/php-debugbar/pull/439
What do you think? I don't think the nonce is need for the non-inline JS files.

barryvdh:
Fixed with #439

PR #479, opened 2021-08-19
Draft: Support twig namespaces

xvilo:
For background information see: https://github.com/maximebf/php-debugbar/issues/478

## What?
- Add a new Collector for Twig that supports their namespace structure.
- Update docs


composer.json
```diff
@@ -18,8 +18,10 @@
     ],
     "require": {
         "php": "^7.1|^8",
+        "ext-json": "*",
         "psr/log": "^1.0",
-        "symfony/var-dumper": "^2.6|^3|^4|^5"
+        "symfony/var-dumper": "^2.6|^3|^4|^5",
+        "twig/twig": "^1.38|^2.7|^3.0"
```
barryvdh, line 24: This should probably be in require-dev of in suggest.

barryvdh:
Is this good? Seems to work

xvilo:
Thanks fo merging @barryvdh we've been using this in our application since PR creation 👍

PR #481, opened 2021-09-28
Add #[ReturnTypeWillChange] attribute to suppress deprecation warning…

rdeprera:
…s and fatal error

Error occurs on php 7.4 >=


glensc:
I'm sure the deprecations happen only >= 8.1, so the PR description saying 7.4 is incorrect

PR #498, opened 2022-01-21
PHP 8.1 fix in DataFormatter.php for "Implicit conversion from float-INF to int loses precision"

beat:
This fixes a PHP 8.1 bug "Deprecated: Implicit conversion from float -INF to int loses precision in vendor/maximebf/debugbar/src/DebugBar/DataFormatter/DataFormatter.php

### DIFF
--- a/src/DebugBar/DataFormatter/DataFormatter.php
+++ b/src/DebugBar/DataFormatter/DataFormatter.php
@@ -78,6 +78,6 @@ class DataFormatter implements DataFormatterInterface
 
         $base = log($size) / log(1024);
         $suffixes = array('B', 'KB', 'MB', 'GB', 'TB');
-        return $sign . round(pow(1024, $base - floor($base)), $precision) . $suffixes[floor($base)];
+        return $sign . round(pow(1024, $base - floor($base)), $precision) . $suffixes[(int) floor($base)];
     }
 }


beat:
Any chance this might get merged this week ? This fix is a no-brainer and compatible from PHP 5 through 8.1.
In PHP, arrays indices can only be text or int type, and by default are int type: Floats are not allowed as keys, before they were auto-converted to ints, but in PHP 8.1 they throw a conversion warning if the float is an rounded integer:
See https://www.php.net/manual/en/language.types.array.php :
> Note:
> Keys that are neither integers nor strings throw a [TypeError](https://www.php.net/manual/en/class.typeerror.php). Such keys can only be generated by a [Traversable](https://www.php.net/manual/en/class.traversable.php) object.

 We plan to release Joomla 4.1 on Feb 15th and would love to have PHP 8.1 compatibility also in the debugbar by then, would it be possible to review and merge this simple fix, so that composer takes it up, @maximebf  ? Merci beaucoup d'avance! :-)

beat:
Merci beaucoup @maximebf ! :-)  When do you plan the next release with push to packagist ? ;-)

laoneo:
@maximebf is there any plan to ship a new release with the PHP 8.1 fixes?

PR #505, opened 2022-05-29
Fix random instances of SQL queries having params repeated

Aberdeener:
In my application, I noticed that on seemingly random occurrences, parameters would be repeated when replacing them in the SQL query string: (notice that they display fine in the table under the query, but the first one is inserted in all 3 locations)
Screen Shot 2022-05-29 at 17 26 38

This PR fixes this issue.

Looking at open issues here, this seems to be the same issue as https://github.com/maximebf/php-debugbar/issues/457, but this problem arises for me without using `BETWEEN` statements... 🤔 

I have added a test, and all tests are passing.


barryvdh:
Thanks!

Aberdeener:
Any idea when a new release will be pushed out with this change in it @barryvdh? 🙏🏼

PR #507, opened 2022-06-10
Fix PHP 8.2 deprecation

jnoordsij:
Dynamic properties are deprecated in PHP 8.2 and will be removed in PHP 9.0. This PR fixes a deprecation error on the `DebugBar/DataFormatter/DataFormatter` class where two dynamic name properties are created in the constructor, by explicitly marking them as (public) class properties.

Reference: https://wiki.php.net/rfc/deprecate_dynamic_properties


jnoordsij:
Any update on this?  
PHP 8.2 now has a first RC and is scheduled for release on November 24, 2022.

PR #530, opened 2023-03-09
allow customization of the memory precision

browner12:
currently there is no way to see the raw "peak_usage" bytes in the output. I'd love to have that number available somewhere, but I couldn't figure out how to do that.

this change is the next best thing, as it allows us to override the default precision of "0", so we can get a little more granularity into the memory usage.


src/DebugBar/DataCollector/MemoryCollector.php
```diff
@@ -23,6 +23,18 @@ class MemoryCollector extends DataCollector implements Renderable
 
     protected $peakUsage = 0;
 
+    protected $precision = 0;
```
PaolaRuby, line 26: Default value must be 2
https://github.com/maximebf/php-debugbar/blob/376ff32ee50c2c24cebd16ccb5a30c647e9b3c6d/src/DebugBar/DataFormatter/DataFormatter.php#L74

PaolaRuby:
> it allows us to override the default precision of "0"

I think what you wanted to modify is this code
https://github.com/maximebf/php-debugbar/blob/376ff32ee50c2c24cebd16ccb5a30c647e9b3c6d/src/DebugBar/DataCollector/MemoryCollector.php#L85

browner12:
thanks!

PR #531, opened 2023-03-22
Fix PHP 8.2 deprecations

stefket:
- Use of "parent" in callables is deprecated
- Creation of dynamic properties is deprecated since php 8.2 (used in some test)


parallels999:
ping @barryvdh :)

creativeprojects:
I am still getting the error, but from line 69:

> Deprecated: Use of "parent" in callables is deprecated in /var/www/html/public/vendor/maximebf/debugbar/src/DebugBar/DataCollector/PDO/TraceablePDOStatement.php on line 69

Does it involve the same fix as line 87?

erikn69:
@creativeprojects try #579

PR #536, opened 2023-06-24
Update JavascriptRenderer.php

sambello247:
I don't think an absolute path is needed here. At least this is what works for me.


barryvdh:
Does it matter?

PR #538, opened 2023-07-04
update highlitght.js to v11.7.0

brice-t:
The API is getting deprecated : change calls according to new API


parallels999:
Did you test this??

I think, if you change the name of the file `highlight.pack.js` to `highlight.min.js`, it would no longer work
https://github.com/maximebf/php-debugbar/blob/293d02c78fd74d97e740772b079c33d61b9e90a7/src/DebugBar/JavascriptRenderer.php#L46
Also, i think that the css needs upgrade
https://github.com/maximebf/php-debugbar/blob/293d02c78fd74d97e740772b079c33d61b9e90a7/src/DebugBar/JavascriptRenderer.php#L41

parallels999:
Fixed on https://github.com/maximebf/php-debugbar/pull/547

parallels999:
Already fixed on 

- https://github.com/maximebf/php-debugbar/pull/547

PR #539, opened 2023-07-13
HtmlVariableList support html on keys

parallels999:
This will allow us to easily customize the presentation of the keys


angeljqv:
Great, i need this

PR #542, opened 2023-07-17
Remove xdebug replacements on file path previews

erikn69:
This method would allow showing only relative paths based on `xdebugReplacements`
this can be very helpful in integrations

This PR hides remote server paths on previews

Before 
image
After
image


erikn69:
This PR hides remote server paths on previews

Before 
image
After
image

PR #549, opened 2023-08-18
[bug] Early return when empty file string

erikn69:
When is empty $file, it returns an invalid url

```php
->getXdebugLink('');
//returns ['url' => 'vscode://file/:1', 'ajax' => false];
->getXdebugLink(null);
//throws exception string required null given on php 8.2
```


angeljqv (approved):

erikn69:
@barryvdh hi, ping

PR #558, opened 2023-09-15
(feat) memory usage  data on measure 

erikn69:
This is an optional config, 
allows you to measure memory usage along with time
image
This helped me optimize a method that causes a fatal memory allocated error (_was a collection filter on a foreach_ 😕)

**Usage:**
```php
$timeCollector = new TimeDataCollector();
$timeCollector->showMemoryUsage(); // without this, it works as before
```


parallels999:
@barryvdh hi,  could you check this? thanks

PR #559, opened 2023-09-15
(fix) set memory baseline on MemoryCollector

### LINKED ISSUE #1324: octane(swoole) memory leak
1. Create a new project.

2. Install octane(swoole) and debugbar.

3. Each time the page is refreshed, the memory increases.

4. Set the enabled value to false in config/debugbar.php

5. Each time the page is refreshed, memory is maintained.

### LINKED ISSUE #1434: debug bar showing 19-21mb memory usage on clean install laravel (welcome page)
I have tested it on laravel 8 , 9 and 10. all of them showing 19-21mb usage on welcome page while having fresh installation.
Is this normal or something is wrong with package. I have php v8.1.17 installed. i did composer update as well. but issue remains same. I have watched few installation tutorials. They have less memory usage than mine for same laravel version.

erikn69:
Fixes https://github.com/barryvdh/laravel-debugbar/issues/1324
Fixes https://github.com/barryvdh/laravel-debugbar/issues/1434

This pr allows only the memory peak to be shown with a baseline, this would normalize the differences in the presentation, currently the maximum peak of the process is shown, but only the memory used by the application should be shown

For example, in Windows a fresh app shows 20MB, while the same app in Linux only shows 3MB.
Added `memory_reset_peak_usage` for only check execution peaks

### DIFF
--- a/src/DebugBar/DataCollector/MemoryCollector.php
+++ b/src/DebugBar/DataCollector/MemoryCollector.php
@@ -17,7 +17,16 @@ class MemoryCollector extends DataCollector implements Renderable
 {
     protected $realUsage = false;
 
+    protected $memoryRealStart = 0;
+
+    protected $memoryStart = 0;
+
     protected $peakUsage = 0;
+
+    public function __construct()
+    {
+        $this->updateMemoryBaseline();
+    }
 
     /**
      * Returns whether total allocated memory page size is used instead of actual used memory size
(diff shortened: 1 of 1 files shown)


barryvdh:
In the case of Laravel, wouldn't this discard the memory usage from the process up to the constructor of this class?

erikn69:
> Laravel, wouldn't this discard the memory usage from the process up to the constructor of this class?

Yes, before the service provider.
But still, it would only show the execution peaks within the script
any other solution?

barryvdh:
For the Octane problem, it would be enough to provide a method to reset for subsequent executions, but calling the memory_reset_peak_usage function might have unexpected results for other tools.

erikn69:
Same app
### WITH THIS PR:
**Linux**
image
**Windows**
image

----
### WITHOUT THIS PR:
**Linux**
image
**Windows**
image

On windows memory  goes up, the same app, also for mac

barryvdh:
Yes but the actual process is using the highest variant, right?

barryvdh:
I removed it from the constructor (for now). We can add it to the laravel-debugbar if Octane is running for example.

erikn69:
> the actual process is using the highest variant, right? 

Without `memory_reset_peak_usage` yes, on the peak

PR #579, opened 2023-10-10
Fix PHP 8.2 deprecations

erikn69:
>**Deprecated:** Use of "parent" in callables is deprecated in /maximebf/debugbar/src/DebugBar/DataCollector/PDO/TraceablePDOStatement.php on line 69


creativeprojects:
Yes, that fixed it for me 👍🏻
Thank you very much

PR #591, opened 2024-02-10
Use xdebug links for messages, move to trait

barryvdh:
Try to use the same functionality in more places.

- Use trait for xdebug to easily use in the other collectors
- Add filename/line to output, so we can show just the filename and standard way
- Default styling for the filename

(Screenshot from laravel debugbar)
image


barryvdh:
Same for the views.

PR #593, opened 2024-02-13
Fix demo PDOCollector name, and update it with 72fea42

erikn69:
https://github.com/maximebf/php-debugbar/commit/72fea42726383318e44e80dd94c77f958274ef47


barryvdh:
Hmm was the pdo->database change a breaking change? 🤔

barryvdh:
Perhaps we should revert that.

PR #594, opened 2024-02-13
Revert "Make PDOCollector easy to extend"

barryvdh:
Reverts maximebf/php-debugbar#582 because I'm not sure of the BC break or not here. See #593


parallels999:
@barryvdh hi, could this be in v2?

PR #596, opened 2024-02-13
Fix jquery4 deprecations

barryvdh:
Guess it would be better to remove jQuery but that's a lot more work 😓


parallels999:
Why don't use `highlight.pack.js` minified version?

barryvdh:
Oh couldn't find it that quickly, but I found a build where you select the languages you need. Need to compare them a bit though.

barryvdh:
Did just the deprecations for now in https://github.com/maximebf/php-debugbar/pull/610

PR #601, opened 2024-02-16
Avoid expanded values on MessagesCollector

PaolaRuby:
When `isHtmlVarDumperUsed()` is false, we have this
image
But, when `isHtmlVarDumperUsed()` is true, all data is expanded by default
image
With this change, it is easier to find the log we are looking for, by the file, line, etc
image
Also as an independent method it is possible to make other customizations based on `$message` data, just by extending the class

### DIFF
--- a/src/DebugBar/DataCollector/MessagesCollector.php
+++ b/src/DebugBar/DataCollector/MessagesCollector.php
@@ -45,6 +45,22 @@ class MessagesCollector extends AbstractLogger implements DataCollectorInterface
     }
 
     /**
+     * @param string|null $messageHtml
+     * @param mixed $message
+     *
+     * @return string|null
+     */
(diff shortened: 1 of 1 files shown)


PaolaRuby:
Also, in te past was `'expanded_depth' => 0, `, but seems that it is not working anymore
https://github.com/maximebf/php-debugbar/blob/689ecda8c1e836c98275df0e638a4d9d8cc83a57/src/DebugBar/DataFormatter/DebugBarVarDumper.php#L22
Now, there is `'maxDepth' => 1 // 1 and 160 are the default values for these options,` but `0` is not allowed
https://symfony.com/doc/current/components/var_dumper.html#dumpers

barryvdh:
Tweaking this and making it configurable in https://github.com/php-debugbar/php-debugbar/pull/601

PR #613, opened 2024-03-11
Revert "Keeps only the main debugbar functionality on iframes"

barryvdh:
Reverts maximebf/php-debugbar#606
See https://github.com/barryvdh/laravel-debugbar/issues/1567


erikn69:
Maybe `window.parent != window.top` to prevent loops
https://github.com/maximebf/php-debugbar/blob/d7b954be25b413e4f062247433a9d5b2a5302667/src/DebugBar/Resources/debugbar.js#L427

```js
this.hasParent = window.parent && window.parent != window.top && window.parent.phpdebugbar && window.parent.phpdebugbar != this;
```

barryvdh:
I'll revert this until we have a chance to test this properly

erikn69:
>Maybe `window.parent != window.top` to prevent loops
>I'll revert this until we have a chance to test this properly

I was in the right direction with `window.parent != window.top`

when you click liveware navigator
image

But after `window.parent != window.top`, no more problem
image

PR #615, opened 2024-03-11
Keeps only the main debugbar functionality on iframes

### LINKED ISSUE #614: InternalError: too much recursion for v1.21+
I'm using Laravel with [Orchid](https://github.com/orchidsoftware/platform) admin panel and [Laravel debugbar](https://github.com/barryvdh/laravel-debugbar)

Since last update I've noticed that debugbar dissapears when I navigate to some page - it works on initial load however. At first I thought it something with project code/deps, but digging deeper I found out it was this particular package (which is laravel-debugbar dependency) and more specifically version 1.21+ (I've installed version 1.20.1 and debugbar start working again)

Orchid is using [Hotwire](https://hotwired.dev/) under the hood so I believe it has to be something with it

Снимок экрана от 2024-03-11 18-45-03

erikn69:
`window.parent != window.top` added to prevent infinite loops(SPA, Livewire, Hotwire https://github.com/maximebf/php-debugbar/pull/613#issuecomment-1988807700)

Alternative to https://github.com/maximebf/php-debugbar/pull/452#issuecomment-1936976144

>We needed a way to add them to the debugbar. (a bit like it happens with ajax+headers).
The debugbar is opened in the top frame, and when something happens inside the iframes, the changes are being added to the top-level debugbar.

image
Some iframes may be hidden or small, this PR propagates data from iframes to the top main debugbar and also it prevents size errors caused by small iframes(the inside smaller iframe change web storage like height, that affects the main debugbar)

Closes #614

### DIFF
--- a/demo/iframes/iframe1.php
+++ b/demo/iframes/iframe1.php
@@ -1,0 +1,13 @@
+<?php
+
+include '../bootstrap.php';
+
+$debugbarRenderer->setBaseUrl('../../src/DebugBar/Resources');
+
+$debugbar['messages']->addMessage('I\'m a IFRAME');
+
+render_demo_page(function() {
+?>
+<iframe src="iframe2.php" style="display:none;"></iframe>
+<?php
+});
--- a/demo/iframes/iframe2.php
+++ b/demo/iframes/iframe2.php
@@ -1,0 +1,9 @@
+<?php
+
+include '../bootstrap.php';
+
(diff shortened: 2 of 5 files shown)


erikn69:
Tested on 
[Hotwire](https://hotwired.dev/) => https://github.com/maximebf/php-debugbar/issues/614#issuecomment-1988946203
[Livewire](https://laravel-livewire.com/) => https://github.com/maximebf/php-debugbar/pull/613#issuecomment-1988807700

PR #617, opened 2024-03-12
Capture HTML body for mails

barryvdh:
If we have the HTML, render that.
image

I'm a bit confused about the showDetailed. I think that the intention is probably the same for both. You probably don't want the details if you show the HTML.

### DIFF
--- a/src/DebugBar/Resources/widgets/mails/widget.js
+++ b/src/DebugBar/Resources/widgets/mails/widget.js
@@ -14,30 +14,42 @@
 
         render: function() {
             this.$list = new  PhpDebugBar.Widgets.ListWidget({ itemRenderer: function(li, mail) {
-                $('<span />').addClass(csscls('subject')).text(mail.subject).appendTo(li);
-                $('<span />').addClass(csscls('to')).text(mail.to).appendTo(li);
-                if (mail.body) {
-                    li.click(function() {
-                        var popup = window.open('about:blank', 'Mail Preview', 'width=650,height=440,scrollbars=yes');
-                        var documentToWriteTo = popup.document;
-                        var headers = !mail.headers ? '' : $('<pre style="border: 1px solid #ddd; padding: 5px;" />')
-                            .append($('<code />').text(mail.headers)).prop('outerHTML');
-                        documentToWriteTo.open();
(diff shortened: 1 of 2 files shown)


barryvdh:
Removed the details and just show the body

barryvdh:
Add text version as summary when both exist
image

src/DebugBar/Resources/widgets/mails/widget.js
```diff
@@ -14,30 +14,43 @@
...
-                            headers.hide();
-                        } else {
-                            headers.show();
-                        }
-                    });
-                }
-            }});
+                    $('<span />').addClass(csscls('subject')).text(mail.subject).appendTo(li);
+                    $('<span />').addClass(csscls('to')).text(mail.to).appendTo(li);
+                    if (mail.html || mail.html) {
```
erikn69, line 19: is `mail.html || mail.html` a typo??

src/DebugBar/Resources/widgets/mails/widget.js
```diff
@@ -14,30 +14,43 @@
...
+                        var header = $('<span />').addClass(csscls('filename')).text('');
+                        $('<a title="Mail Preview">View Mail</a>').on('click', function () {
+                            var popup = window.open('about:blank', 'Mail Preview', 'width=650,height=440,scrollbars=yes');
+                            var documentToWriteTo = popup.document;
+                            var headers = !mail.headers ? '' : $('<pre style="border: 1px solid #ddd; padding: 5px;" />')
+                                .append($('<code />').text(mail.headers));
+
+                            var body = $('<pre style="border: 1px solid #ddd; padding: 5px;" />').text(mail.body)
+                            var html = null;
+                            if (mail.html) {
```
erikn69, line 29: after `if (mail.html || mail.html) {`, this is always true

src/DebugBar/Resources/widgets/mails/widget.js
barryvdh: Yeah one should be body

src/DebugBar/Resources/widgets/mails/widget.js
barryvdh: Yes see other comment, should be when body is set but not html

PR #619, opened 2024-03-12
Fix html body mail on SwiftMailCollector

parallels999:
- Issue introduced on #617

After #617, SwiftMailCollector stop showing messages body preview,
Now `html` array key is needed

https://github.com/maximebf/php-debugbar/blob/0b407703b08ea0cf6ebc61e267cc96ff7000911b/src/DebugBar/Bridge/Symfony/SymfonyMailCollector.php#L65-L66


barryvdh:
Hmm maybe we should check if the key exists

barryvdh:
Is the body actually html here? If so, the other body can be null or the text version

barryvdh:
I think this was already fixed with https://github.com/maximebf/php-debugbar/pull/618 though, can you confirm?

barryvdh:
If it's html, we need to add it to html, otherwise to the body; https://github.com/maximebf/php-debugbar/pull/620

PR #620, opened 2024-03-12
Swiftmailer: Set html to null

barryvdh:


parallels999:
This is not showing html preview

parallels999:
Better set body to null

PR #622, opened 2024-03-13
Fix connection filter on sqls widget

erikn69:
Each new ajax page could bring new connections
The connections accumulate, and appear in all records, even if no SQL exists with that connection

----
Fix a css problem
image

image

### DIFF
--- a/src/DebugBar/Resources/widgets/sqlqueries/widget.js
+++ b/src/DebugBar/Resources/widgets/sqlqueries/widget.js
@@ -14,15 +14,7 @@
 
         onFilterClick: function(el) {
             $(el).toggleClass(csscls('excluded'));
-
-            var excludedLabels = [];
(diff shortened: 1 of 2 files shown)


src/DebugBar/Resources/widgets/sqlqueries/widget.js
```diff
@@ -14,15 +14,7 @@
 
         onFilterClick: function(el) {
             $(el).toggleClass(csscls('excluded'));
-
-            var excludedLabels = [];
-            this.$toolbar.find(csscls('.filter') + csscls('.excluded')).each(function() {
-                excludedLabels.push(this.rel);
-            });
-
```
erikn69, line 22: I did remove this code because it does nothing, seems like a copy/paste issue 
https://github.com/maximebf/php-debugbar/blob/1fdc3a3007561721daec958a1a75d7a062d1e6ab/src/DebugBar/Resources/widgets.js#L438-L449

Issue #598, opened 2024-02-14
[Request] Remove font-awesome


barryvdh:
Based on https://github.com/maximebf/php-debugbar/pull/272 
Can we just use https://tabler.io/icons icons as svg inline? Either add the default ones to css, or allow the collectors to define svg (base64 encoded or something?)

parallels999:
Seems like a big change, possibly a breaking change, a lot of icons, and dependencies use custom Indicators with fa icons, they would lose the icons

barryvdh:
Yeah might warrant a 2.x version.

miken32:
As additional motivation for removing this, I'm seeing [this notice](https://github.com/FortAwesome/Font-Awesome/issues/19925) popping up repeatedly on my console lately. Fixed in FA 6.5, but a very old version is bundled in to this code.

PR #624, opened 2024-03-13
Make font-awesome css a single file

parallels999:
Make font-awesome css a single file using base64, 
This makes it easier to extend `DebugBar\JavascriptRenderer`, without this i get `file not found` on browser
At least until it is removed #598


barryvdh:
So this is just the woff that base64 encoded right? Guess we don't need the rest anway.

barryvdh:
Ah yes this is what I do with laravel-debugbar also; https://github.com/barryvdh/laravel-debugbar/blob/master/src/Resources/vendor/font-awesome/style.css

barryvdh:
Yeah indeed. Not sure why I didn't just do that here in the first place as well though :P

barryvdh:
Done in https://github.com/barryvdh/laravel-debugbar/pull/1596

PR #631, opened 2024-03-20
Update highlightjs to custom 11.9

barryvdh:
Downloaded with https://highlightjs.org/download
Languages: php, javascript, sql, shell, css, plaintext, xml

Maybe the github.css needs to be updated, but I think we're already running a custom style.


barryvdh:
Size went from 120kB to 64kB so that's nice.

parallels999:
It would have been 54kb if you had minified it

barryvdh:
Hmm I think I just copied the minified file from the bundle 🤔

PR #636, opened 2024-03-21
Better iframe support

erikn69:
Tested on Livewire, Hotwire
`window.parent` can cause problems in many scenarios, 
`window.top` would only work with iframes [(how-to-check-a-webpage-is-loaded-inside-an-iframe-or-into-the-browser)](https://www.geeksforgeeks.org/how-to-check-a-webpage-is-loaded-inside-an-iframe-or-into-the-browser-window-using-javascript/)
This is a better option
image

### DIFF
--- a/demo/iframes/iframe2.php
+++ b/demo/iframes/iframe2.php
@@ -6,4 +6,14 @@ $debugbarRenderer->setBaseUrl('../../src/DebugBar/Resources');
 
 $debugbar['messages']->addMessage('I\'m a Deeper Hidden Iframe');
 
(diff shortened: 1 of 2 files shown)


barryvdh:
> @barryvdh hi, was the iframes support removed? or did it just stop working due to a bug?
> 
> image
> Now each iframe displays its own debugbar, and the data in hidden/small iframes is inaccessible.

Fixed in https://github.com/php-debugbar/php-debugbar/pull/956
Was because it was in a scope, not attached to window.

PR #637, opened 2024-03-21
Fix label on dataset tab

erikn69:
The label is not appearing when I select an item from the request history
image


erikn69:
@barryvdh what means the badge??
Always show the requests count, means last request nb
image

barryvdh:
Yes shows the number of requests

erikn69:
is this a CSS problem or a expected behaivor?
image

barryvdh:
Hmm that is a css issue. Those badges belong to the rows of the request. Is this with a default install?

erikn69:
> Is this with a default install?

Testing master branchs, I wanted to see how that new data set is going to work

PR #641, opened 2024-03-31
Refactor browser test, test monolog bridge

barryvdh:


barryvdh:
Not sure if I'm going to do all, but it's a decent start I guess :)

PR #644, opened 2024-04-02
use font-awesome icons for close/open

barryvdh:
It doesn't really change much, but cleaner and easier to style (background-color) in themes (eg for Laravel debugbar)
image


barryvdh:
I merged #645 in this and added a text-align to achieve the same effect.
image

Not sure if this is better though, we could just use the svg's now that they are centered, and override the fill.
The only downside would be for the case that Font-Awesome is not available for some reason.

barryvdh:
Don't think we need this for now.

Issue #555, opened 2023-09-04
How to exclude paths like `vendor/` on ExceptionsCollector traces


angeljqv:
Is there a way to exclude/hide traces on ExceptionsCollector? 
If it doesn't exist the option, can it be a feature?

PR #645, opened 2024-04-02
Improve align of icons in the header on the right side

nckrtl:
This PR adds proper alignment of the icons on the right side, before and after:
before-after

What I've done:
- Extract the svg's and fix alignment in the view box. All svg content was not exactly centered within the view box. I fixed that in Figma, exported them and url encoded the SVG code. Now all icons don't have to be centered with arbitrary values like 6px 6px, instead they are now centered with the `center` value;
- Applied proper hex code of `#555` to improve uniformity
- Set fix sizes + flex box props to properly align the history icon
- Removed padding from the minimise/maximise icons

There was also a line of 1 pixel high separating some elements in header bar  from the header bar bottom border `div.phpdebugbar-header > div > *` this was caused by float: left making the elements being targeted by this css to become 16.31px instead of 16px. Setting a fixed height fixes this.

I also had to adjust the padding on the widgets list item to nicely align them with the close icon:
Screenshot 2024-04-02 at 23 57 19

This PR might come across as being overkill on the spacing and stuff but its part of an extraction of the changes presented here: https://github.com/barryvdh/laravel-debugbar/pull/1601


erikn69:
It seems like you would have conflicts with #644

nckrtl:
Partially I think, as my spacing change are still valid. But indeed best to wait for that one to get merged and I'll update this one.

barryvdh:
The open buttons looks empty. I think it was missing the viewbox, which I added in https://github.com/maximebf/php-debugbar/commit/100030444436d9f7f549aed29c92def790245035
(You can test this in the demo by enabling the open handler in the demo/bootstrap.php

```
$debugbar = new StandardDebugBar();
$debugbarRenderer = $debugbar->getJavascriptRenderer()
                             ->setBaseUrl('../src/DebugBar/Resources')
                             ->setAjaxHandlerEnableTab(true)
                             ->setOpenHandlerUrl('/')
                             ->setEnableJqueryNoConflict(false);
```

barryvdh:
Followup in https://github.com/barryvdh/laravel-debugbar/pull/1603 which replaces the laravel icons (only change the fill in dark mode)

PR #647, opened 2024-04-04
UI refinements toolbar

nckrtl:
This PR add come UI refinements to the toolbar:
- Better readability of the counter in the tabs by making the background black on selection
- Better vertical alignment of the counter
- The tooltip now resizes based on the text thanks to white-space no wrap. This solves the big "padding" on the sides
- Adjusted background positioning of the logo using `center` value instead of arbitrary values. This change make it seem the logo got replace but that isn't the case
- Tooltips are better readabile when displayed in front of text on the page by apply a backdrop blur

Tested these changes in Safari, Chrome and Firefox.

ui-refinements


barryvdh:
Why did you change the border-radius of the badge?

barryvdh:
Hmm I think we can keep the old one for now.

barryvdh:
Hmm I think the #000 might be a bit 'sharp'. What do you think of #555 as active badge color
image

nckrtl:
Yeah that'll work as well, still high enough contrast.

PR #658, opened 2024-06-18
Fix not rendering correctly in some cases

matsuo:
Thank you for your developing!

I found the debug bar is not displayed correctly when the query string contains "<!--<script>" (confirmed with maximebf/debugbar v1.22.3).
This pull request fixes the issue. Please check it out.

Pattern 1: sample query string
```
?<!--<script>
```

Pattern 2: sample code
```
$debugbar["messages"]->addMessage("<!--<script>");
```


angeljqv:
LGTM 👍

PR #666, opened 2024-08-20
Reduce memory usage for consecutive requests with stacked data

### LINKED ISSUE #665: Exponential memory usage growth with stacked data and Request collector
The request data collector store the data from `$_SESSION` which contains the stacked data, which in turn contains the data from the request collector, etc, so memory usage double with each request. So with multiple requests that have a request data collector and getting stacked, that becomes a problem rather quickly.

dotdash:
Currently, the session stores stacked data, which in turn might contain the session data if a RequestCollector is active. This leads to quadratic growth in memory usage and rapid performance degradation if multiple requests add stacked data before the next time the debug bar is rendered.

We can avoid this be temporarily removing the stacked data from the session while collection data.

Fixes #665


dotdash:
Ping :) Any chance that someone could have a look at this?

Knoph1:
Hi.
Can I help out please?

Knoph1:
A;; seen thanks.

PR #671, opened 2024-09-07
Fix parameter output if parameter is Datetime or array [v1.22.4]

aliusa:
Version: [v1.22.4](https://github.com/maximebf/php-debugbar/releases/tag/v1.22.4)
Change causing crash: Escape doctrine entries by @barryvdh in https://github.com/maximebf/php-debugbar/pull/657

Fixes error:
```php
Error (E_ERROR):	htmlentities(): Argument #1 ($string) must be of type string, DateTime given
File:	/.../vendor/maximebf/debugbar/src/DebugBar/Bridge/DoctrineCollector.php
Line:	87
```

If passing Datetime instance to Doctrine ORM QueryBuilder.
```php
$doctrineOrmQueryBuilder
    ->andWhere("item.date_from <= :date_from")
    ->setParameter('date_from', new \DateTime($array['date_from']), Types::DATE_MUTABLE);
```

Also fixes argument output if array is given.

image

### DIFF
--- a/src/DebugBar/Bridge/DoctrineCollector.php
+++ b/src/DebugBar/Bridge/DoctrineCollector.php
@@ -84,7 +84,15 @@ class DoctrineCollector extends DataCollector implements Renderable, AssetProvid
     {
         $params = [];
         foreach ($query['params'] ?? [] as $name => $param) {
-            $params[$name] = htmlentities($param?:"", ENT_QUOTES, 'UTF-8', false);
+            if (is_string($param) || is_numeric($param)) {
+                $params[$name] = htmlentities($param, ENT_QUOTES, 'UTF-8', false);
+            } elseif ($param instanceof \DateTimeInterface) {
+                $params[$name] = $param->format('Y-m-d H:i:s');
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks! Added testcase here: https://github.com/maximebf/php-debugbar/pull/673

phpcodemonkey:
@aliusa @barryvdh This PR introduces a breaking change - it seems to be when the params contain a nested array - the implode on line 89 throws an exception 'Array to string conversion' - so the solution provided seems over-simplistic to the type of parameter shapes that can be received. v1.22.4/1.22.5 do not cause the problem.

barryvdh:
Should be fixed with https://github.com/maximebf/php-debugbar/releases/tag/v1.23.1 (and https://github.com/maximebf/php-debugbar/pull/681)

phpcodemonkey:
> Should be fixed with https://github.com/maximebf/php-debugbar/releases/tag/v1.23.1 (and #681)

Wow, that was fast! Thank you! 🙏 😄 

(I hadn't seen the other issue number! 🙄 )

PR #672, opened 2024-09-08
Hide empty collectors until used

barryvdh:
This will hide empty collectors, until the collector is actually filled using any data.
This can be useful for less-used collectors without filling up te screen.

Note:
 - It is not hidden when empty, to keep a more consistent flow when using ajax etc.
 - it's not using the badge because not every collector uses a badge

Not sure if it's worth creating a config option for.

### DIFF
--- a/src/DebugBar/JavascriptRenderer.php
+++ b/src/DebugBar/JavascriptRenderer.php
@@ -62,7 +62,7 @@ class JavascriptRenderer
 
     protected $useRequireJs = false;
 
-    protected $hideEmptyTabs = false;
+    protected $hideEmptyTabs = null;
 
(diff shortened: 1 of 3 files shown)


barryvdh:
Or should we have a threshold for the collectors, so we can make it configurable?

barryvdh:
Added a config flag

parallels999:
I don't know if it's a bug, but when I'm in a tab, and an ajax comes with the same tab but empty, 
the data from the old tab remains open, shouldn't It show another tab that's not hidden?

Example, I sent a completely empty ajax, all the tabs were hidden, but the `database` tab remained open with the `index` data
image

PR #676, opened 2024-09-10
Enable copy, tweak icon

barryvdh:
Instead of the alert, toggle the copy button with a similar icon for 2 seconds, to indicate success.
image


erikn69:
great 🎉🎉, it was very annoying

barryvdh:
I know, right :p 
Never enabled the button but really bugged me when I tried.

erikn69:
but now I see that laravel-debugbar has its own file, this change will not be reflected there
And copy button was placed inline and the query end
image

barryvdh:
Yes I was going to fix it in laravel-debugar in the new release, but first here.

PR #699, opened 2024-12-15
Check if hljs has language

barryvdh:
For example the Laravel error page has hljs now, which doesn't contain SQL. So that will fail otherwise.


josevv28:
Hi, It seems the releases are no longer shown in https://packagist.org/packages/maximebf/debugbar
could it be because of `maximebf/php-debugbar` to `php-debugbar/php-debugbar/` change?

barryvdh:
Sorry webhooks are still a bit wonky. I thinks they are okay now but updated manually now.

PR #701, opened 2024-12-27
Revert dragging

barryvdh:
I tested this, and I'm not 100% sure about it. I like the idea but it feels really like and edge case. I can see the appeal of moving the debugbar to the right instead of left, but I think that should be either configurable in the UI or as option.

The dragging has a bit of a downside of keeping the draggable state when switching tabs or something. Not really sure but it doesn't feel reliable yet.

### DIFF
--- a/src/DebugBar/Resources/debugbar.css
+++ b/src/DebugBar/Resources/debugbar.css
@@ -106,13 +106,6 @@ a.phpdebugbar-restore-btn {
   border-right: 1px solid #ddd;
 }
 
-.phpdebugbar-dragging a.phpdebugbar-restore-btn {
-  cursor: move;
-  cursor: grabbing;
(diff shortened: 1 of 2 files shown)


parallels999:
I just add css, 
```css
div.phpdebugbar-closed{
  right:0;
  left:auto !important;
}
```

barryvdh:
Yeah might make that an option or something

barryvdh:
> I just add css,
> 
> ```css
> div.phpdebugbar-closed{
>   right:0;
>   left:auto !important;
> }
> ```

Did something like that in https://github.com/php-debugbar/php-debugbar/pull/703 which makes it a UI setting and switches the border.
Screenshot 2025-01-14 at 11 45 14

PR #703, opened 2025-01-02
Dark Theme + Theme variables + UI tweaks

barryvdh:
Initial POC
Screenshot 2025-01-02 at 19 27 40@2x
Screenshot 2025-01-02 at 19 28 03@2x

This basically just replaces the hardcoded background/text/borders with variables and adds a switcher.

You can set the theme to auto to match the browser, or light/dark to enabled it.

Probably needs a lot of overrides still.


barryvdh:
Alright signing off for the weekend, think this covered most of it.
Screenshot 2025-01-02 at 21 27 48


let me know if you spot something or have some improvements!

barryvdh:
Theme switcher: 
Screenshot 2025-01-06 at 14 47 33

Not extremely happy with it, but perhaps a good starting point for more options (like left/right position)

barryvdh:
dark
light
minimized

angeljqv:
Another idea: what if tabs could dynamically add their options to the settings panel?

For example, `PDOCollector` could set a time threshold, and queries exceeding that threshold would be considered slow and highlighted in red, without that setting, none would be highlighted.

angeljqv:
Could the editor be selected in the settings tab?

PR #721, opened 2025-02-07
Support html messages on MessagesCollector

erikn69:
From https://github.com/barryvdh/laravel-debugbar/issues/1654#issuecomment-2637599284
>Unfortunately, no URLs allowed
>```php
>Debugbar::getCollector('Quick Links')->addMessage('<a href="https://google.com">Test</a>');
>```
image

----
**Now:**
```php
Debugbar::getCollector('Quick Links')->addMessage('<a href="https://google.com">Test</a>', 'links', false);
```
image

### DIFF
--- a/src/DebugBar/DataCollector/MessagesCollector.php
+++ b/src/DebugBar/DataCollector/MessagesCollector.php
@@ -71,7 +71,7 @@ class MessagesCollector extends AbstractLogger implements DataCollectorInterface
     public function addMessage($message, $label = 'info', $isString = true)
     {
         $messageText = $message;
-        $messageHtml = null;
+        $messageHtml = !$isString ? $message : null;
         if (!is_string($message)) {
             // Send both text and HTML representations; the text version is used for searches
             $messageText = $this->getDataFormatter()->formatVar($message);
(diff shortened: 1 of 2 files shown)


barryvdh:
Wouldn't this lead to possible XSS? Although the parameter should be given explicitly.

erikn69:
The third argument must be sent as `false`, by default it is `true`, basically you have to force the printing of html, also this package should only be used for development, not in production, but if it is a security risk it would be better to discard this PR

erikn69:
You could also use the third argument to indicate that it should be formatted to a specific lang for highlightjs,
here is an example indicating `xml`
```php
->addMessage("<?xml version='1.0' encoding='UTF-8'?>\n<iva>\n\t<TypoIDInfor>R</TypoIDInfor>\n\t<IdInfor>0000</IdInfor>\n\t<rSocial>BM SA</rSocial>\n\t<year>2025</year>\n\t<month>01</month>\n</iva>", 'data', 'xml');
```
image

barryvdh:
I'm not really sure about this. This leaves potentially missed exploits. What is your exact use case?

PR #722, opened 2025-02-07
Fix warning on null query parameter

UrGuardian4ngel:
Fix PHP 8 warning in `TracedStatement::getSqlWithParameters()` when null was bound

Since PHP 8.0, [built-in functions like `strtr()` will emit a warning when passing `null` to required string parameters](docs):

```
strtr(): Passing null to parameter #1 ($string) of type string is deprecated
```

This can happen when e.g. binding `PDO::PARAM_NULL` to your query.

[docs]: https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal

### DIFF
--- a/src/DebugBar/DataCollector/PDO/TracedStatement.php
+++ b/src/DebugBar/DataCollector/PDO/TracedStatement.php
@@ -114,9 +114,12 @@ class TracedStatement
 
         foreach ($this->parameters as $k => $v) {
 
-            $backRefSafeV = strtr($v, $cleanBackRefCharMap);
-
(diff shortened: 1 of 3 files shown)


UrGuardian4ngel:
Code was based on latest `v1.23.5` tag.
Not sure which branch I should target for `v1.x`?

barryvdh:
Master if fine, the fix can be in the 2.x version.

barryvdh:
I backported this to https://github.com/php-debugbar/php-debugbar/releases/tag/v1.23.6

PR #723, opened 2025-02-13
Allow dataset to defer loading instead of passing all the data

barryvdh:
Gives the option to defer the loading of the dataset. This gives the change for a request to write more data (eg. terminable middleware) before loading, and make the requests smaller. 

Next step would be to only load or render when the debugbar is open..


cesarreyes3:
This could serve to #726, this way you could automatically delete the non-ajax call files, I also think it would speed up the loading of the page if the data is loaded separately(if there is too much data)

PR #728, opened 2025-02-17
Fix defer data loading

erikn69:
I tried to test this functionality and I always get that the file does not exist, and in the sources the file is saved when doing `getData()`, with this change it works for me, or am I doing something wrong?
https://github.com/php-debugbar/php-debugbar/blob/917a032fbab522e0b82232a12fd1eeb439020919/src/DebugBar/DebugBar.php#L248-L252

Ref #723

### DIFF
--- a/src/DebugBar/JavascriptRenderer.php
+++ b/src/DebugBar/JavascriptRenderer.php
@@ -1148,6 +1148,7 @@ class JavascriptRenderer
 
         $suffix = !$initialize ? '(ajax)' : null;
         if ($this->areDatasetsDeferred()) {
+            $this->debugBar->getData();
             $js .= $this->getLoadDatasetCode($this->debugBar->getCurrentRequestId(), $suffix);
         } else {
             $js .= $this->getAddDatasetCode($this->debugBar->getCurrentRequestId(), $this->debugBar->getData(), $suffix);


barryvdh:
Hmm I think it would be better to defer this as late as possible, but not sure how

cesarreyes3:
With this it works as I expected

My storage 
```php
class TempFileStorage extends \DebugBar\Storage\FileStorage 
{
    public function get($id)
    {
        $data = parent::get($id);
        try {
            unlink($this->makeFilename($id));
        } catch(\Throwable $e){}

        return $data;
    }
}
```
My debugbar
```php
$myDebugbar->setStorage(new TempFileStorage(__DIR__ . '/debugbar-temp'));
$myDebugbar->getJavascriptRenderer()
    ->setAjaxHandlerAutoShow(true)
    ->setAjaxHandlerEnableTab(true)
    ->setDeferDatasets(true);
```

cesarreyes3:
It would be great if there was some configuration that automatically detects if the call is ajax and sends the request id in the header without having to add `$debugbar->sendDataInHeaders(true);` after any ajax calls

or am I overlooking something?

https://github.com/php-debugbar/php-debugbar/blob/917a032fbab522e0b82232a12fd1eeb439020919/demo/ajax.php#L4-L7

cesarreyes3:
ping @barryvdh

PR #730, opened 2025-02-17
Update JavascriptRenderer.php

### LINKED ISSUE #729: Hardcoded paths to old installation path
[JavascriptRenderer.php](https://github.com/php-debugbar/php-debugbar/blob/master/src/DebugBar/JavascriptRenderer.php#L105C24-L105C75) has hard coded paths to `'/vendor/maximebf/...`

Causes this:
Image

Image

jefhar:
Update path to reflect new installation.

Fixes #729

### DIFF
--- a/src/DebugBar/JavascriptRenderer.php
+++ b/src/DebugBar/JavascriptRenderer.php
@@ -101,15 +101,19 @@ class JavascriptRenderer
     {
         $this->debugBar = $debugBar;
 
-        if ($baseUrl === null) {
-            $baseUrl = '/vendor/php-debugbar/php-debugbar/src/DebugBar/Resources';
-        }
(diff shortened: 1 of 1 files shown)


barryvdh:
Can we check both paths? The old name still works.

parallels999:
maybe check which of the two directories exists

barryvdh:
Or use the magic __DIR__ constant

jefhar:
@barryvdh only path exists when the `maximebf/debugbar` package isn't installed. Does composer have a nifty feature that will install a single repository based on 2 different names?

barryvdh:
I checked the basePath to find the vendor folder. Should work for both cases, but fall back to the string.

Although I would really recommend not having the vendor folder accessible in the public root.

PR #735, opened 2025-02-20
Avoid LF, \t, multiple spaces on timeline

erikn69:
This prevents it from becoming illegible in the cases described

Before this change
image

After this change
image


barryvdh:
Heh I just had this issue this afternoon, thanks

PR #746, opened 2025-02-28
build: replace maximebf/debugbar in composer.json

JaZo:
I'm using a plugin for the debugbar that still requires `maximebf/debugbar`, so now I have two versions installed with the same classes. By using the replace, we can make sure only the new package name is installed.


JaZo:
For anyone interested; as a (temporary) workaround you can add this to your own `composer.json` to make sure the old package isn't installed:

```json
"replace": {
    "maximebf/debugbar": "*"
},
```

PR #748, opened 2025-03-04
OPT-IN Start collecting warnings, notices and deprecations on ExceptionsCollector

cesarreyes3:
It would be great to be able to collect the other types of errors that don't stop the script from running.

**Example:**
```php
$cadena = "";
$cadena--;
```
image

### DIFF
--- a/src/DebugBar/DataCollector/ExceptionsCollector.php
+++ b/src/DebugBar/DataCollector/ExceptionsCollector.php
@@ -22,6 +22,44 @@ class ExceptionsCollector extends DataCollector implements Renderable
     protected $chainExceptions = false;
 
     /**
+     * Start collecting warnings, notices and deprecations
+     */
+    public function collectWarnings() {
+        $self = $this;
+        $errorTypes = array(
+            1    => 'E_ERROR',
+            2    => 'E_WARNING',
+            4    => 'E_PARSE',
+            8    => 'E_NOTICE',
+            16   => 'E_CORE_ERROR',
+            32   => 'E_CORE_WARNING',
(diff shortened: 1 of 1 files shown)


barryvdh:
Wouldn't this conflict with the existing error handler?

cesarreyes3:
Hi @barryvdh, In the case of Vanilla PHP does not seem to me, but on some framework like laravel, yes

Here is another aproach:
```php
public function collectWarnings($preserveOriginalHandler = true) {
        ///....
        $originalHandler = $preserveOriginalHandler ? set_error_handler(null) : null; // get previus handler hack
        set_error_handler(function ($errno, $errstr, $errfile, $errline) use ($self, $errorTypes, $originalHandler) {
            $self->exceptions[] = array( // don't know whether to use $self or just $this, please correct me
                'type' => $errorTypes[$errno] ?? 'UNKNOWN',
                'message' => $errstr,
                'code' => $errno,
                'file' => $self->normalizeFilePath($errfile),
                'line' => $errline,
                'stack_trace' => null,
                'stack_trace_html' => null,
                'surrounding_lines' => null,
                'xdebug_link' => $self->getXdebugLink($errfile, $errline)
            );

            if ($originalHandler) {
                return call_user_func($oldHandler, $errno, $errstr, $errfile, $errline);
            }

            return false;
    });

}
```
What do you think @barryvdh, would it work?? **Working Demo:** https://onlinephp.io/c/4214a

cesarreyes3:
Seems to work on vanilla php
image
Seems to work on laravel
image

PR #749, opened 2025-03-04
ExceptionsCollector easy to extend

cesarreyes3:
In case #748 doesn't get merged, i could extend the class with that code and just return
```php
public function collect() {
    return array(
        'count' => count($this->exceptions),
        'exceptions' => $this->exceptions
    );
}
```
Also for anyone who wants to extend the same class, just with another name


barryvdh:
Is this still required?

cesarreyes3:
After the other PR is merged, it is not required by me, but it does not affect anything if someone wants to use it in the future.

PR #750, opened 2025-03-04
Remove ExceptionsCollector deprecations on V2

cesarreyes3:
Many deprecations were left when going from v1 to v2
[php-debugbar deprecations](https://github.com/search?q=repo%3Aphp-debugbar%2Fphp-debugbar%20depreca&type=code)


barryvdh:
Whoops to late now unfortunately..

PR #751, opened 2025-03-07
Compact symfony var dumper, and better text

cesarreyes3:
It's easier to read when there is more than one exception. Most of the time it is not necessary to see all the traces

Before 
image

After
image


parallels999:
it works, and with `Ctrl + Click` you can open all the items 👍

image

PR #770, opened 2025-04-21
Support doctrine 3.x

### LINKED ISSUE #768: Doctrine integration outdated
The current doctrine collector requires a DebugStack, this isnt in the latest version of Doctrine/DBAL which now uses middleware rather than logging stacks, the only other method of getting it to work is with an entityManager, which is provided with the ORM, but im only using the DBAL

### LINKED ISSUE #774: setSQLLogger is deprecated
$entityManager->getConnection()->getConfiguration()->setSQLLogger($debugStack);

is depreceated right? any plans to fix it?

erikn69:
- Closes #768
- Closes #774

https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/configuration.html

The code was moved from `\DebugBar\Bridge\DoctrineCollector` to `\DebugBar\Bridge\Doctrine\DoctrineCollector` and the middleware was added, it keeps 2.x support on `\DebugBar\Bridge\DoctrineCollector`

the middleware wraps doctrine connection, is there a better way to achieve this?

**UPDATE1:** it seems that doctrine 3.x does not support php 8.0
**UPDATE2:** `durationBackground` added, events like `transaction` are now better differentiated, events are avoided from being counted as `statements`


barryvdh:
Would it make more sense to create php-debugbar/doctrine-bridge or something?

barryvdh:
It's a bit difficult to pin dependencies and things like that. And making changes or upgrading in the future with breaking changes, make it also difficult.

erikn69:
>Would it make more sense to create php-debugbar/doctrine-bridge or something?

Maybe, but since there was already support for 2x I thought about adding it.
In that case, shouldn't all the bridges be separated? Monolog for example

>It's a bit difficult to pin dependencies and things like that. And making changes or upgrading in the future with breaking changes, make it also difficult.

Is it because of the integration test? Yes, it's difficult.

barryvdh:
well yeah it's mostly difficult because we don't actually add the depencies so we cannot bump anything.

erikn69:
In my opinion, the only thing that saves this type of thing is that it is not a production package, you can adapt it to your local environment in your own way.

erikn69:
>New way is not worse performance-wise right?

That shouldn't be the case, but try it and tell us, I don't use Doctrine, I'm just trying to help

barryvdh:
I think I'm gonna make a seperate repo. Just thinking about how to do it cleanly. Shall I make the base package and let @erikn69 PR his code (copy paste) it?

erikn69:
Closed by [php-debugbar/doctrine-bridge/releases/v2.0.0-beta1](https://github.com/php-debugbar/doctrine-bridge/releases/tag/v2.0.0-beta1)

PR #773, opened 2025-04-29
bugfix(redis-storage): Unserialise is sometimes false

ClaraLeigh:
This is likely indicative of a deeper issue with the data being fetched but when there is a problem with unserialize, we should fallback to an empty array when the unserialize function returns false.


parallels999 (approved):

adnedelcu:
Any idea when this will be released? We've been encountering this issue ourselves and it's pretty annoying.

We don't have repro steps, but we think it happens when the request is very big (for example including a big XML string as part of the body for a POST request).

PR #807, opened 2025-11-25
Fix CSP nonce on JSON encoded script tags

### LINKED ISSUE #806: CSP nonce not added to JSON encoded strings containing <script>
This is a follow on from #563 which already describes the situation well so I'll just quote the short version.

> This PR fixes csp errors caused by <script> tags without a nonce generated by HtmlDumper.

However, #658 changed the level of escaping when JSON encoding so `'<script>'` becomes `'\\u003Cscript\\u003E'`.

PeterMead:
Fixes #806


barryvdh:
Additionally: https://github.com/php-debugbar/php-debugbar/pull/849

PR #690, opened 2024-10-21
Pretty-printing SQL queries button(Indentation formatter)

erikn69:
PoC of [sql-formatter-org/sql-formatter](https://github.com/sql-formatter-org/sql-formatter) from https://github.com/barryvdh/laravel-debugbar/issues/1654#issuecomment-2427142327

image
image

Based on:
https://github.com/maximebf/php-debugbar/blob/689720d724c771ac4add859056744b7b3f2406da/src/DebugBar/Resources/widgets/mails/widget.js#L21-L23

PR #861, opened 2026-01-01
Add sql expand

barryvdh:
Screenshot 2026-01-01 at 20 54 26@2x

Fixes https://github.com/php-debugbar/php-debugbar/pull/690


resources/widgets/sqlqueries/widget.js
```diff
@@ -265,6 +265,14 @@
                             return '';
                         }
                         table.hidden = !table.hidden;
+                        const code = li.querySelector('code');
+                        if (code && typeof phpdebugbar_sqlformatter !== 'undefined') {
```
erikn69: Should it work in laravel-debugbar too?
Do I need to activate any settings there?

because `phpdebugbar_sqlformatter` always returns undefined

erikn69: I can see it now, thanks

PR #938, opened 2026-01-13
Check if xhr/fetch are already bound before wrapping

barryvdh:
Fixes https://github.com/fruitcake/laravel-debugbar/issues/1882#issuecomment-3743871114

Avoids binding when the fetch/xhr is already wrapped by us.


barryvdh:
or nevermind, I'll create a testcase and fix

PR #956, opened 2026-01-20
Fix iframe support

barryvdh:


erikn69:
I noticed that the variable name can be changed through the settings and I have created another alternative.

demo/iframes/iframe1.php
```diff
@@ -2,7 +2,11 @@
 
 include '../bootstrap.php';
 
-$debugbarRenderer->setBaseUrl('../../resources');
+/** @var \DebugBar\DebugBar $debugbar */
+/** @var \DebugBar\JavascriptRenderer $debugbarRenderer */
+
+$debugbarRenderer->setAssetHandlerUrl('../assets.php')
+    ->setOpenHandlerUrl('../open.php');
```
erikn69: You can avoid this the way I do it; wouldn't that be better?

barryvdh: It would, but I'm using Laravel herd and that renders /demo in the subfolder, so wouldn't work there.

erikn69:
I tried your solution, the demo doesn't work 100%,
the AJAX call inside `iframe2.php` isn't being captured

My PR
image

This PR
image

barryvdh:
Was that because of your fix in the debugbar or in the settimeout in the demo?

PR #979, opened 2026-01-29
ui: add index to executed queries list

sobhanlori:
Add index column to query list for better readability

When many queries are executed, it’s difficult to reference
a specific query. This change adds a sequential index to each
query row to improve clarity and debugging experience.


barryvdh:
Hmm I'm not really convinced how that makes it more clear.

PR #982, opened 2026-02-05
fix: QueryFormatter::quoteBinding() $binding must be of type string, false given

calebdw:
Hello!

When the json_encode() function fails, it returns false which causes a TypeError downstream. Instead, we should throw an exception so that the actual cause can be inspected.

In this case, the real error that I was experiencing is json_encode() erroring with `Malformed UTF-8 characters, possibly incorrectly encoded`

Thanks!

### DIFF
--- a/src/DataFormatter/QueryFormatter.php
+++ b/src/DataFormatter/QueryFormatter.php
@@ -35,7 +35,20 @@ class QueryFormatter extends DataFormatter
             }
 
             if (is_object($binding)) {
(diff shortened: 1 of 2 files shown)


calebdw:
Possibly, but in general the jeon_encode call should trow an exception instead of returning false and causing a type error faw away

Either that or the false case needs to be properly handled

barryvdh:
Is it possible to add a test for something that failes? So the closure + something thats not json_encodable?

PR #983, opened 2026-02-05
Add mergeMeasures feature to TimeDataCollector

### LINKED ISSUE #434: Aggregate timeline measures and keep their time positions
I follow up #287 pull-request about aggregate timeline. This is an awesome enhancement, but I think we could go further.

Why not merging this two Timeline systems in one. 
- First timeline do not aggregate same event measures but give them a position in the timeline.
- Second timeline system aggregates same event measures, but we lose time position information

Here is a possible implementation on Roadiz CMS:
image

I do not stack measures to keep timing location and ease up debugging when you need to know when a event repeats itself.

erikn69:
Closes #434

PoC, Merge repeated measures into a single timeline entry, reusing the same segment instead of creating new ones

image

### DIFF
--- a/resources/widgets.js
+++ b/resources/widgets.js
@@ -787,7 +787,7 @@
                             aggregate[group] = { count: 0, duration: 0, memory: 0 };
                         }
 
-                        aggregate[group].count += 1;
+                        aggregate[group].count += values.length;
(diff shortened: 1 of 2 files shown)


barryvdh:
Thanks!

PR #984, opened 2026-02-06
Handle invalid UTF-8 characters in JSON encoding

### LINKED ISSUE #1958: Malformed JSON
On v4 I'm getting frequently TypeErrors from https://github.com/php-debugbar/php-debugbar/blob/f52f0f148a6cc32b4a16c00ce74f369b81bea378/src/Storage/FileStorage.php#L57 stating that `null` is returned when `array` is expected. It's because `json_encode` is writing `""` (`false`) to the file from https://github.com/php-debugbar/php-debugbar/blob/f52f0f148a6cc32b4a16c00ce74f369b81bea378/src/Storage/FileStorage.php#L37

`json_encode` is failing with `Malformed UTF-8 characters, possibly incorrectly encoded`. The malformed data is coming from the `QueryCollector` `bindings`. There are a couple of queries where I'm storing binary data which is returned by `inet_pton()`.

I think adding `JSON_INVALID_UTF8_SUBSTITUTE` to `json_encode` is probably the only fix? That will replace the binary data with �

If you're happy with that approach I will send a PR.

erikn69:
Closes https://github.com/fruitcake/laravel-debugbar/issues/1958
```php
\Debugbar::log("Text with fix " . "\xC3\xC3\xC3\xC3\xC3" . " end");
```

Before:
image
```
TypeError: DebugBar\Storage\FileStorage::get(): Return value must be of type array, null returned
```

**After:**
image

### DIFF
--- a/src/Storage/FileStorage.php
+++ b/src/Storage/FileStorage.php
@@ -34,7 +34,7 @@ class FileStorage extends AbstractStorage
             mkdir($this->dirname, 0o755, true);
             file_put_contents($this->dirname . '.gitignore', "*\n!.gitignore\n");
         }
-        file_put_contents($this->makeFilename($id), json_encode($data));
+        file_put_contents($this->makeFilename($id), json_encode($data, JSON_INVALID_UTF8_SUBSTITUTE));
 
         $this->autoPrune();
     }
--- a/src/Storage/FileStorage.php
+++ b/src/Storage/FileStorage.php
@@ -54,7 +54,7 @@ class FileStorage extends AbstractStorage
             throw new \RuntimeException("Unable to read file $fileName");
         }
 
-        return json_decode($content, true);
+        return json_decode($content, true) ?: [];
     }
 
     /**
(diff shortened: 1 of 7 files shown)


barryvdh:
Hmm so I had a fix for this in php-debugbar, but it was removed in laravel-debugbar. Fixed in https://github.com/fruitcake/laravel-debugbar/pull/1959

Adding some extra checks in https://github.com/php-debugbar/php-debugbar/pull/985

src/Storage/FileStorage.php
```diff
@@ -54,7 +54,7 @@ public function get(string $id): array
             throw new \RuntimeException("Unable to read file $fileName");
         }
 
-        return json_decode($content, true);
+        return json_decode($content, true) ?: [];
```
erikn69: @barryvdh you missed some checks

barryvdh: https://github.com/php-debugbar/php-debugbar/pull/986

PR #987, opened 2026-02-10
Recalculate body padding after SPA navigation

angus-mcritchie:
## Problem

When using SPA navigation frameworks (Livewire, Turbo, HTMX), the body's padding may change between pages. The debugbar captures body padding at initialization time (lines 702-704) and uses these cached values in `recomputeBottomOffset()`.

This causes a visual bug when navigating between pages with different layouts. For example, navigating from a page with a sticky header (large `padding-top` like `177px`) to a page without one (no padding), the debugbar incorrectly restores the old padding value after navigation completes.

### Steps to reproduce (using Livewire 4)

1. Create two pages with different body padding (e.g., one with a sticky header, one without)
2. Enable debugbar with `bodyBottomInset: true` and `toolbarPosition: 'top'`
3. Load the page with the sticky header (body has `padding-top: 177px`)
4. Use `wire:navigate` to navigate to the page without the sticky header
5. Observe: The debugbar restores `padding-top: 177px` even though the new page has no padding

### Timeline of the bug

```
Page A loads (sticky header) → debugbar caches bodyPaddingTopHeight = 177
User navigates to Page B (no header) → Livewire clears body padding
livewire:navigated fires → debugbar calls recomputeBottomOffset()
recomputeBottomOffset() restores paddingTop = 177px (wrong!)
```

## Solution

This PR adds listeners for common SPA navigation events and recalculates the cached body padding values after navigation completes:

- `livewire:navigated` (Livewire 4)
- `turbo:load` (Turbo/Hotwire)
- `htmx:afterSettle` (HTMX)

The new `recalculateBodyPadding()` method:
1. Clears any debugbar-applied inline padding
2. Reads the page's actual CSS padding values
3. Updates the cached values
4. Calls `recomputeBottomOffset()` with the correct values

## Testing

Tested with Livewire 4 wire:navigate between pages with different body padding values. The debugbar now correctly respects each page's actual padding.

### DIFF
--- a/tests/Tests/JavascriptRendererTest.php
+++ b/tests/Tests/JavascriptRendererTest.php
@@ -55,6 +55,7 @@ class JavascriptRendererTest extends DebugBarTestCase
             'ajax_handler_enable_tab' => true,
             'defer_datasets' => true,
             'csp_nonce' => 'mynonce',
+            'spa_navigation_events' => ['custom:event'],
         ]);
 
         $this->assertEquals('/foo', $this->r->getBasePath());
@@ -80,6 +81,7 @@ class JavascriptRendererTest extends DebugBarTestCase
         $this->assertTrue($this->r->isAjaxHandlerTabEnabled());
         $this->assertTrue($this->r->areDatasetsDeferred());
         $this->assertEquals('mynonce', $this->r->getCspNonce());
+        $this->assertEquals(['custom:event'], $this->r->getSpaNavigationEvents());
     }
 
     public function testAddAssets(): void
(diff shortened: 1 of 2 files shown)


barryvdh:
I think this makes sense, but should we make this configurable? Maybe set a list of eventNames somewhere? Otherwise every framework needs to add their own event.

angus-mcritchie:
> Looks good :) Can you fix the tests to add the new initialization code?

Sure, all done!

Big fan of your work BTW, the debugbar has saved so much time and it was actually one of the reasons I wanted to try Laravel back in the day.

barryvdh:
Thanks!

PR #989, opened 2026-02-18
PdoStorage.php

thirdwheel:
Allow the use of SQL wildcards in filters except utime

### DIFF
--- a/src/Storage/PdoStorage.php
+++ b/src/Storage/PdoStorage.php
@@ -91,7 +91,14 @@ class PdoStorage extends AbstractStorage
 
         foreach ($filters as $key => $value) {
             if (in_array($key, ['datetime', 'uri', 'ip', 'method'], true)) {
-                $where[] = "meta_$key = ?";
+                if (strpos($value, '%') !== false)
+                {
+                    $where[] = "meta_$key LIKE ?";
+                }
+                else
+                {
+                    $where[] = "meta_$key = ?";
+                }
                 $params[] = $value;
             } elseif ($key === 'utime') {
                 $where[] = "meta_utime > ?";


barryvdh:
Hmm, the filestorage uses https://www.php.net/manual/en/function.fnmatch.php
I think we should probably use the same semantics.

datetime -> not sure how you would filter, I think this might need the same behavior as utime?

method -> should just be a match, not really usefull to search patterns

So we should search for the patterns for uri and ip

https://3v4l.org/eEDAG#vnull

chatGPT suggest something like this


```php
function globToMysqlLike(string $pattern, string $escapeChar = '\\'): string
{
    // Escape LIKE special characters first
    $pattern = str_replace(
        [$escapeChar, '%', '_'],
        [$escapeChar . $escapeChar, $escapeChar . '%', $escapeChar . '_'],
        $pattern
    );

    // Convert glob wildcards to SQL wildcards
    return str_replace(['*', '?'], ['%', '_'], $pattern);
}

foreach ($filters as $key => $value) {
    if (in_array($key, ['uri', 'ip', 'method'], true)) {

        // Only switch to LIKE if pattern chars are present
        if (strpbrk($value, '*?') !== false) {
            $where[] = "meta_$key LIKE ? ESCAPE '\\\\'";
            $params[] = globToMysqlLike($value);
        } else {
            $where[] = "meta_$key = ?";
            $params[] = $value;
        }

    } elseif (in_array($key, ['datetime', 'utime'], true)) {
        $where[] = "meta_$key > ?";
        $params[] = $value;
    }
}
```

thirdwheel:
Yeah I realised after submitting this that the other storage types had the ability to search, using fnmatch() as you said. Leave it with me, I'll come up with something without relying on AI.

PR #990, opened 2026-02-19
PdoStorage.php

thirdwheel:
Add `fnmatch()` style wildcards to PDO backend for `uri` and `ip` filters


src/Storage/PdoStorage.php
```diff
@@ -173,6 +176,22 @@ public function setSqlQueries(array $queries): void
         $this->sqlQueries = array_merge($this->sqlQueries, $queries);
     }
 
+    public function globToSql(string $pattern, string $escapeChar = '\\')
```
barryvdh:
```suggestion
    public function globToSql(string $pattern, string $escapeChar = '\\'): string
```

PR #995, opened 2026-02-19
Bug Report: `isIframe` detection incorrectly uses namespace object instead of instance

jeffersongoncalves:
**Affects:**
- `php-debugbar/php-debugbar` v3.4.0 (and likely earlier)
- `fruitcake/laravel-debugbar` v4.0.9

---

## Summary

The `isIframe` detection in `resources/debugbar.js` tests `window.top.PhpDebugBar`, which
resolves to the **PhpDebugBar namespace object** — not a DebugBar instance. Since any object
is truthy in JavaScript, `isIframe` is always set to a truthy value whenever
`window.self !== window.top`, causing the debugbar to hide itself even when it is the main
page (not a child frame).

---

## Bug Location

**File:** `resources/debugbar.js`
**Line:** 708

```js
// BUGGY — evaluates to the namespace object (truthy), not a boolean
this.isIframe = window.self !== window.top && window.top.PhpDebugBar && window.top.PhpDebugBar;
```

---

## Root Cause

`window.top.PhpDebugBar` points to the `PhpDebugBar` **namespace/class object** declared by
the debugbar assets:

```js
// From JavascriptRenderer.php output
var PhpDebugBar = PhpDebugBar || {};
PhpDebugBar.DebugBar = class DebugBar { ... };
```

The intent is to detect whether the parent frame has an **active debugbar instance**. For
that, `JavascriptRenderer.php` already assigns the instance to a dedicated property:

```js
// JavascriptRenderer.php — line 1238
window.PhpDebugBar.instance = phpdebugbar;
```

The bug is that the check uses `window.top.PhpDebugBar` (the class namespace, always defined
after the assets load) instead of `window.top.PhpDebugBar.instance` (the actual running
instance, only present when the debugbar has rendered in the parent frame).

---

## Impact

When `window.self !== window.top` is `true` (e.g. the page loads inside any frame), both
consequences fire:

1. **`render()` hides the element:**
   ```js
   render() {
       if (this.isIframe) {
           this.el.hidden = true; // debugbar becomes invisible
       }
       // ...
   }
   ```

2. **`restoreState()` returns early without showing the toolbar:**
   ```js
   restoreState() {
       if (this.isIframe) {
           return; // state is never restored, toolbar stays hidden
       }
       // ...
   }
   ```

Result: the debugbar toolbar is completely invisible to the user.

---

## Fix

Replace `window.top.PhpDebugBar` with `window.top.PhpDebugBar?.instance` so the check only
returns truthy when a real DebugBar instance exists in the parent frame:

```js
// FIXED
try {
    this.isIframe = window.self !== window.top && !!(window.top.PhpDebugBar?.instance);
} catch (_error) {
    this.isIframe = false;
}
```

The optional chaining (`?.`) handles the case where the parent frame exists but has not loaded
the debugbar assets yet, avoiding a `TypeError`. The `!!` cast ensures the result is always a
proper boolean.

---

### `php-debugbar/php-debugbar`

| File | Line | Change |
|------|------|--------|
| `resources/debugbar.js` | 708 | Replace the buggy expression (see diff below) |
| `resources/dist/debugbar.min.js` | — | Rebuild from `resources/debugbar.js` after fix |

**Diff (`resources/debugbar.js`):**

```diff
 try {
-    this.isIframe = window.self !== window.top && window.top.PhpDebugBar && window.top.PhpDebugBar;
+    this.isIframe = window.self !== window.top && !!(window.top.PhpDebugBar?.instance);
 } catch (_error) {
     this.isIframe = false;
 }
```

### `fruitcake/laravel-debugbar`

No JavaScript changes are needed here — it depends on `php-debugbar/php-debugbar` for the
JS assets. After the upstream fix is released, bump the minimum version constraint in
`composer.json`:

```json
"require": {
    "php-debugbar/php-debugbar": "^3.4.1"
}
```

---

## Verification

After applying the fix, the debugbar should:

1. Render visibly on pages that are not inside a frame (`window.self === window.top`).
2. Still correctly detect and hide itself when running as a child frame where the parent
   window has an active debugbar instance (`window.top.PhpDebugBar.instance` is set).

---

## Workaround (until upstream is patched)

In `AppServiceProvider::boot()` (or equivalent), listen for the `RequestHandled` event and
inject a script that corrects `isIframe` and re-runs `restoreState()` after the debugbar
initialises:

```php
use Illuminate\Foundation\Http\Events\RequestHandled;
use Illuminate\Support\Facades\Event;

Event::listen(RequestHandled::class, function (RequestHandled $event): void {
    $content = $event->response->getContent();

    if (! $content || ! str_contains($content, 'phpdebugbar')) {
        return;
    }

    $fix = <<<'JS'
        <script>
        document.addEventListener('DOMContentLoaded', function () {
            if (typeof phpdebugbar === 'undefined') return;
            phpdebugbar.isIframe = (window !== window.top);
            if (!phpdebugbar.isIframe) {
                phpdebugbar.el.hidden = false;
                phpdebugbar.restoreState();
            }
        });
        </script>
        JS;

    $event->response->setContent(str_replace('</body>', $fix . '</body>', $content));
});
```

This runs after the debugbar's own `DOMContentLoaded` listener (its `<script>` is earlier in
the HTML), so it correctly overrides `isIframe`, unhides the element, and calls
`restoreState()` again.


barryvdh:
Hmm why would the namespace be active when the Debugbar is not used? Instance can be checked but it will be later so might not be ready at the same time of the iframe?

PR #1063, opened 2026-04-29
fix: duplicate highlight disappears after sorting

erikn69:
- Bug added on #863

After using the sort button, the highlighting of duplicate queries disappears, as does the functionality to show only duplicates.

If you use the `Show only duplicates` link and then sort, all data is displayed, but the link remains with `Show All`

### DIFF
--- a/resources/widgets/sqlqueries/widget.js
+++ b/resources/widgets/sqlqueries/widget.js
@@ -271,6 +271,38 @@
             this.list = new PhpDebugBar.Widgets.ListWidget({
                 itemRenderer: (li, stmt) => this.itemRenderer(li, stmt)
             });
+            this.list.bindAttr('data', function (data) {
+                const sql = {};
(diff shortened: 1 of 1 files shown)


erikn69:
I noticed something strange in the JS assets; in the queries widget, both the compressed and uncompressed versions are listed simultaneously, but only the uncompressed version is executed. They don't appear to cause any conflicts, but it seems like a bug.
The problem seems to be in laravel-debugbar.

barryvdh:
> I noticed something strange in the JS assets; in the queries widget, both the compressed and uncompressed versions are listed simultaneously, but only the uncompressed version is executed. They don't appear to cause any conflicts, but it seems like a bug. The problem seems to be in laravel-debugbar.

Good spot! Fixed in https://github.com/php-debugbar/php-debugbar/pull/1065

PR #1069, opened 2026-06-22
Handle mail body fallback to HTML content on Mail Text Version(mail widget)

erikn69:
When the email is in HTML format, the text version is always empty.
Is it possible to show the raw HTML as the text plain version?

image


barryvdh:
I'm not sure what clients show instead?

erikn69:
After this change:
image

PR #1072, opened 2026-07-26
fix(query-formatter): handle boolean SQL bindings

oleksandr-kurychyk:
Handle boolean values separately when formatting SQL queries with bindings.

Previously, boolean bindings were passed directly to quoteBinding(), which accepts only string values. With strict types enabled, this caused the following TypeError:

QueryFormatter::quoteBinding(): Argument #1 ($binding) must be of type string, bool given

Convert boolean values to SQL literals:

true → TRUE
false → FALSE

This prevents formatter errors and produces a valid, readable SQL representation for boolean bindings.


barryvdh:
Thanks!

PR #1073, opened 2026-09-14
fix(time): allow measures of the same name to nest

maicol07:
`startedMeasures` is keyed by name only. Start a measure while one with the same name is already running and the pending entry is overwritten. The inner `stopMeasure()` then removes it, and the enclosing one throws `Failed stopping measure '$name' because it hasn't been started`.

This bites whenever instrumented code re-enters itself. The case I hit: a Laravel app using [`inertia-modal`](https://github.com/emargareten/inertia-modal), which renders a modal's base route through a nested `Router::dispatch()`. That fires a second `PreparingResponse`/`ResponsePrepared` pair, and `laravel-debugbar`'s `TimeCollectorProvider` starts and stops `'Preparing Response'` on those events, so the outer stop throws and the request 500s. A recursive `measure()` call on the same closure hits it too.

The fix keeps a per-name stack. `stopMeasure()` pops the most recently started entry, the enclosing measure stays pending until its own stop call, and `collect()` drains everything left over. Both measures land on the timeline with correct durations.

I also considered guarding with `hasStartedMeasure()` in `laravel-debugbar`'s `TimeCollectorProvider`, the way fruitcake/laravel-debugbar#1964 and fruitcake/laravel-debugbar#1966 were fixed over there. It stops the crash but drops one of the two measures and leaves the outer one's start time wrong. The re-entrancy is real, so it seemed better handled here.

Two things worth flagging. `$startedMeasures` is `protected`, so the shape change is visible to subclasses. I found nothing in this repo or the bridges touching it directly, but I can't speak for downstream code. And with `mergeRepeatedMeasures()` on, nested same-name measures merge into one entry with their durations summed, so the nested time gets counted twice. That's how merge already treats repeated measures, and I left it alone rather than special-case containment. Tell me if you'd rather I handle it in this PR.

_Note: PR contents and description generated with Claude Opus 5_


src/DataCollector/TimeDataCollector.php
```diff
@@ -95,7 +96,7 @@ public function startMeasure(string $name, ?string $label = null, ?string $colle
      */
     public function hasStartedMeasure(string $name): bool
     {
-        return isset($this->startedMeasures[$name]);
+        return ! empty($this->startedMeasures[$name]);
```
erikn69:
```suggestion
        return ! empty($this->startedMeasures[$name] ?? null);
```
This method is public; it could be called with a name of a measure that doesn't exist.

maicol07, line 99: This shouldn't be needed. Afaik, `empty` doesn't trigger any exception when the array key is missing

